PR #15, opened 2022-10-05
feat: added new field for selection interface

smsochneg:
solving problem when instanceof doesn't work in webpack production build for selection

### DIFF
--- a/src/selection.d.ts
+++ b/src/selection.d.ts
@@ -1,0 +1,7 @@
+import prosemirrorState from 'prosemirror-state';
+
+declare module 'prosemirror-state' {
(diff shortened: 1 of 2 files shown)


src/selection.d.ts
```diff
@@ -0,0 +1,7 @@
+import prosemirrorState from 'prosemirror-state';
+
+declare module 'prosemirror-state' {
+    class Selection extends prosemirrorState.Selection {
+        selectionName?: string;
```
d3m1d0v, line 5: must be a static field

d3m1d0v (approved):
Pls, add a description of what it's for and what bug fixes it

PR #21, opened 2022-10-11
feat: switching table rows when pressing up and down arrow keys on keyboard

smsochneg:
Navigation between table rows with keyboard arrow keys

### DIFF
--- a/src/table-utils/utils.ts
+++ b/src/table-utils/utils.ts
@@ -1,5 +1,7 @@
 import {Node as PmNode} from 'prosemirror-model';
 import {findChildren, findParentNode, Predicate} from 'prosemirror-utils';
+import type {EditorView} from 'prosemirror-view';
+import {isTextSelection} from '../utils/selection';
 import {TableRole} from './const';
 
 export const isTableNode: Predicate = (node) => node.type.spec.tableRole === TableRole.Table;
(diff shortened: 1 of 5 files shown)


d3m1d0v (changes requested):
You need to handle cells with multiple rows of content. Now if a table cell contains multiple paragraphs and the cursor is in the first paragraph and I press down key, the cursor will move to the next row, but should move to the next paragraph inside the current cell

src/table-utils/commands/goToNextRow.ts
```diff
@@ -0,0 +1,70 @@
...
+            index = dir < 0 ? $head.index(d) : $head.indexAfter(d);
+        if (index !== (dir < 0 ? 0 : parent.childCount)) return null;
+        if (parent.type.spec.tableRole === 'tableCell') {
+            const cellPos = $head.before(d);
+            const dirStr = dir > 0 ? 'down' : 'up';
+            return view.endOfTextblock(dirStr) ? cellPos : null;
+        }
+    }
+    return null;
+}
```
d3m1d0v, line 21: is this copy of the same function from `table-utils/utils`?

src/table-utils/utils.ts
```diff
@@ -1,5 +1,7 @@
 import {Node as PmNode} from 'prosemirror-model';
+import {TextSelection} from 'prosemirror-state';
 import {findChildren, findParentNode, Predicate} from 'prosemirror-utils';
+import {EditorView} from 'prosemirror-view';
```
d3m1d0v, line 4:
```suggestion
import type {EditorView} from 'prosemirror-view';
```

src/table-utils/utils.ts
```diff
@@ -28,3 +30,19 @@ export const getTableDimensions = (node: PmNode | Node) => {
...
 };
+
+export function atEndOfCell(view: EditorView, dir: number) {
+    if (!(view.state.selection instanceof TextSelection)) return null;
+    const {$head} = view.state.selection;
+    for (let d = $head.depth - 1; d >= 0; d--) {
+        const parent = $head.node(d),
+            index = dir < 0 ? $head.index(d) : $head.indexAfter(d);
+        if (index !== (dir < 0 ? 0 : parent.childCount)) return null;
+        if (parent.type.spec.tableRole === 'tableCell') {
```
d3m1d0v, line 41: Use value from enum: `parent.type.spec.tableRole === TableRole.Cell`

src/table-utils/utils.ts
```diff
@@ -28,3 +30,19 @@ export const getTableDimensions = (node: PmNode | Node) => {
 
     return {rows: rows || 1, cols: cols || 1};
 };
+
+export function atEndOfCell(view: EditorView, dir: number) {
+    if (!(view.state.selection instanceof TextSelection)) return null;
```
d3m1d0v, line 35: you should use `isTextSelection` helper from `src/utils/selection` instead of `instanceof`

d3m1d0v (approved):

PR #26, opened 2022-10-27
feat(Lists): breaking the list when deleting a list item

d3m1d0v:
Lifting a list item and breaking the list when the cursor is at the beginning of the list item and the Backspace button has been pressed

### DIFF
--- a/src/extensions/markdown/Lists/commands.ts
+++ b/src/extensions/markdown/Lists/commands.ts
@@ -1,7 +1,9 @@
-import type {NodeType} from 'prosemirror-model';
-import type {Command} from 'prosemirror-state';
-import {wrapInList} from 'prosemirror-schema-list';
-import {findAnyParentList} from './utils';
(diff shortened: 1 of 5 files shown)


src/extensions/markdown/Lists/commands.ts
```diff
@@ -15,3 +17,60 @@ export function toList(listType: NodeType): Command {
         return wrapInList(listType)(state, dispatch);
     };
 }
+
+export const liftIfCursorIsAtBeginningOfItem: Command = (state, dispatch, view) => {
+    const $cursor = get$Cursor(state.selection);
+    if (!$cursor) return false;
+    const {schema} = state;
+    const parentBlock = $cursor.node($cursor.depth - 1);
+    console.log('liftIfCursorIsAtBeginningOfItem cursors', $cursor.pos);
```
smsochneg: excess console.log

smsochneg (approved):

PR #187, opened 2024-02-14
fix(deps): add markdown-it-attrs to deps

d3m1d0v:
`markdown-it-attrs` is in deps in `@diplodoc/transform`

But we use attrs plugin directly in _core_, so it should be in deps

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "@types/is-number": "^7.0.1",
     "@types/markdown-it": "^12.2.3",
     "is-number": "^7.0.0",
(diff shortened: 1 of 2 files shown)


smsochneg (approved):

package.json
```diff
@@ -99,6 +99,7 @@
     "@types/is-number": "^7.0.1",
     "@types/markdown-it": "^12.2.3",
     "is-number": "^7.0.0",
+    "markdown-it-attrs": "4.1.4",
```
ogonkov, line 102: Why version is fixed?

PR #236, opened 2024-05-13
cm-home-end

burashka:


d3m1d0v:
not relevant after the codemirror update

PR #237, opened 2024-05-16
Update extensions to behaviour

makhnatkin:
1. Moved the parts related to yfm from behavior to yfm
2. Fixed some eslint errors


src/react-utils/useNodeHovered.ts
```diff
@@ -6,6 +6,7 @@ export const useNodeHovered = (nodeRef: RefObject<HTMLElement>) => {
     const [nodeHovered, setNodeHovered, unsetNodeHovered] = useBooleanState(false);
 
     useEffect(() => {
+        // TODO: @makhnatkin is used current?
```
makhnatkin, line 9: @d3m1d0v Should we change the dependencies from nodeRef to nodeRef.current for the warning fix if there is no ESLint: React House Effect has a missing dependency: 'nodeRef'. Either include it or remove the dependency array. Mytable values like 'nodeRef.current' aren't valid dependencies because mutating them doesn't re-render the component.(react-hooks/exhaustive-deps) ?

src/extensions/markdown/CodeBlock/CodeBlockHighlight/CodeBlockHighlight.ts
```diff
@@ -11,11 +11,12 @@ import {Step} from 'prosemirror-transform';
...
-import {ExtensionAuto} from '../../../core';
-import {codeBlockLangAttr, codeBlockNodeName, codeBlockType} from '../../../extensions/markdown';
-import {capitalize} from '../../../lodash';
-import {logger} from '../../../logger';
-import {codeLangSelectTooltipViewCreator} from '../CodeMirrorView/TooltipPlugin';
+import {ExtensionAuto} from '../../../../core';
+import {capitalize} from '../../../../lodash';
+import {logger} from '../../../../logger';
+// TODO: check cycle imports
+import {codeLangSelectTooltipViewCreator} from '../../../behavior/CodeMirrorView/TooltipPlugin';
```
d3m1d0v, line 18: We need to move or copy TooltipPlugin from CodeMirrorView to CodeBlockHighlight extension, because CodeMirrorView extension well be removed soon

d3m1d0v (approved):

PR #248, opened 2024-05-30
feat(note, tabs): added leave block logic for second enter

makhnatkin:


src/extensions/yfm/YfmNote/commands.ts
```diff
@@ -31,6 +31,37 @@ export const exitFromNoteTitle: Command = (state, dispatch) => {
...
+        isSameNodeType($cursor.node(-2), noteType(schema)) &&
+        $cursor.node(-1).childCount > 1
+    ) {
+        // current texblock is last child
+        if ($cursor.after() === $cursor.end(-1)) {
+            const nodeRange = new NodeRange(
+                state.doc.resolve($cursor.before()),
+                state.doc.resolve($cursor.after()),
+                $cursor.depth - 1,
+            );
```
d3m1d0v, line 54: Can we replace this with `$cursor.blockRange()!` like in YfmCut?

src/extensions/yfm/YfmTabs/plugins.ts
```diff
@@ -218,6 +219,40 @@ export const tabEnter: Command = (state) => {
...
+
+    if (
+        $cursor.depth > 2 && // depth must be at least 3
+        isSameNodeType($cursor.node(-1), tabPanelType(schema)) &&
+        isSameNodeType($cursor.node(-2), tabsType(schema))
+    ) {
+        // current texblock is last child
+        if ($cursor.after() === $cursor.end(-1)) {
+            if (dispatch) {
+                const copedNode = $cursor.parent.copy();
```
d3m1d0v, line 239:
```suggestion
                const copiedNode = $cursor.parent.copy();
```

src/extensions/yfm/YfmNote/commands.ts
makhnatkin: fixed

PR #250, opened 2024-05-31
ci: setup github actions

d3m1d0v:


.github/workflows/ci.yml
```diff
@@ -4,29 +4,40 @@ on:
   push:
     branches: [master]
```
makhnatkin, line 5: Let's switch to main

d3m1d0v: yeah, got it

makhnatkin (approved):

PR #303, opened 2024-07-26
feat: support YfmLangOptions drilling to markup editor

saxumcordis:


src/markup/codemirror/create.ts
```diff
@@ -112,7 +114,7 @@ export function createCodemirror(params: CreateCodemirrorParams) {
             ...defaultKeymap,
             ...historyKeymap,
         ]),
-        autocompletion(),
+        autocompletion(autocompletionConfig),
         yfmLang(),
```
makhnatkin, line 118: We [added an option](https://github.com/gravity-ui/markdown-editor/pull/315/files#diff-3918eac2ddd197df4e93220958532f747d2ef57c4da6432a565c0737af7db140R138) to the yfmHtml plugin. 

Now you can add your function to autocomplete things. For example, look at how `mdAutocomplete` [works](https://github.com/gravity-ui/markdown-editor/pull/315/files#diff-3918eac2ddd197df4e93220958532f747d2ef57c4da6432a565c0737af7db140R88).

For example you can write a function for suggest with `CompletionsSource` interface from `@codemirror/autocomplete` and add it to options in `yfmLang`

src/markup/codemirror/create.ts
```diff
@@ -112,7 +114,7 @@ export function createCodemirror(params: CreateCodemirrorParams) {
             ...defaultKeymap,
             ...historyKeymap,
         ]),
-        autocompletion(),
```
makhnatkin, line 115: Let us revert to the original code; it would seem that there is no longer a need for overriding options.

src/markup/codemirror/create.ts
```diff
@@ -53,6 +53,7 @@ export type CreateCodemirrorParams = {
     uploadHandler?: FileUploadHandler;
     needImgDimms?: boolean;
     extraMarkupExtensions?: Extension[];
+    autocompletionConfig?: Parameters<typeof autocompletion>[0];
```
makhnatkin, line 56: Let's change this prop – we will forward the functions for the autocomplete, not the configuration of the plugin

src/markup/codemirror/create.ts
saxumcordis: great!

makhnatkin (approved):

PR #313, opened 2024-07-30
fix(bundle): scroll to current line after moving cursor in markup mode

d3m1d0v:


demo/Playground.tsx
```diff
@@ -218,64 +219,73 @@ export const Playground = React.memo<PlaygroundProps>((props) => {
...
+                    />
+                    <DropdownMenu.Item
+                        text="Move to line"
+                        action={() => {
+                            mdEditor.moveCursor({line: 115});
+                            mdEditor.focus();
+                        }}
+                    />
+                </DropdownMenu>
+                <MoveToLine
```
makhnatkin, line 281: Should we show this field in the wysiwyg mode?

makhnatkin:
now it looks like a label for an input
I suggest swapping the button and the input field, as well as making the button a non-flat style,
Screenshot 2024-07-30 at 13 26 12

d3m1d0v:
> now it looks like a label for an input I suggest swapping the button and the input field, as well as making the button a non-flat style, Screenshot 2024-07-30 at 13 26 12

done it!

makhnatkin (approved):

PR #331, opened 2024-08-14
feat: support overriding codemirror history extension

beliarh:


src/bundle/Editor.ts
```diff
@@ -111,6 +111,10 @@ type ChangeEditorModeOptions = {
 export type MarkupConfig = {
     /** Additional extensions for codemirror instance. */
     extensions?: CreateCodemirrorParams['extensions'];
+    /** Can be used to override default history extension */
+    historyExtension?: CreateCodemirrorParams['historyExtension'];
+    /** Can be used to override default history keymap */
+    historyKeymap?: CreateCodemirrorParams['historyKeymap'];
```
makhnatkin, line 117: Now we have created 2 properties specifically for one particular extension, which may be a temporary solution.

In general, we should come up with a mechanism for connecting extensions and keymap for them

makhnatkin (approved):

PR #350, opened 2024-08-29
feat(bundle): scroll to top of screen when scrolling to line in markup mode

d3m1d0v:


makhnatkin (approved):

src/bundle/Editor.ts
```diff
@@ -509,3 +531,17 @@ export class EditorImpl extends SafeEventEmitter<EventMapInt> implements EditorI
         return serializedEditorMarkup?.trim() !== wysiwygValue.trim();
     }
 }
+
+function calculateCSSNumberValue(cssValue: string): number {
```
makhnatkin, line 535: Let's add a comment that this is necessary in case there is an expression in the ccc variable in order to put it in the helper in the future

PR #354, opened 2024-09-03
feat(yfmHtmlBlock): updated yfm-html-block

makhnatkin:


src/extensions/yfm/YfmHtmlBlock/YfmHtmlBlockNodeView/YfmHtmlBlockView.tsx
```diff
@@ -130,17 +138,35 @@ const YfmHtmlBlockPreview: React.FC<YfmHtmlBlockViewProps> = ({html, onСlick, c
         }
     };
 
+    // finds all relative links (href^="#") and changes their click behavior
+    const createAnchorLinkHandlers = (type: 'add' | 'remove') => () => {
```
makhnatkin, line 142: The code is increasingly being duplicated with runtime. I think in future versions, we should switch to calling the script at runtime for previewing in shift mode

d3m1d0v (approved):

PR #356, opened 2024-09-05
feat: support colspan and rowspan in table actions

d3m1d0v:


makhnatkin (approved):

src/table-utils/table-desc.ts
```diff
@@ -0,0 +1,190 @@
...
+
+        // table –> tbody -> ...
+        const baseOffset = 2;
+
+        const rows: TableRowDesc[] = [];
+        tbody.forEach((trow, offset) => {
+            rows.push({node: trow, offset, cells: []});
+        });
+
+        tbody.forEach((trow, _1, trowIndex) => {
```
makhnatkin, line 52: Perhaps the problem can be solved in `O(N)` - if you go through the cells and put metadata about them in a hash map.

At the same time, I assume that the keys of the HashMap will contain only those cell numbers that have colspan or rowspan attributes, so most often, the size of this object in memory will not be maximal.

makhnatkin (approved):

PR #360, opened 2024-09-11
Add i18n for Table's and Mermaid's context menu

St1ggy:


src/i18n/yfm-table/ru.json
```diff
@@ -5,5 +5,14 @@
   "row.add.before": "Добавить строку до",
   "row.add.after": "Добавить строку после",
   "row.remove": "Удалить строку",
-  "table.remove": "Удалить таблицу"
+  "table.remove": "Удалить таблицу",
```
makhnatkin, line 8: "table.remove" and "table.menu.column.remove" are same
Maybe we should leave one key?

makhnatkin (approved):

Issue #368, opened 2024-09-16
Sticky behavior breaks in editor with parent overflow: auto


makhnatkin:
#### Problem Statement
When the editor is placed inside a parent container with `overflow: auto`, the condition at [this line](https://github.com/gravity-ui/markdown-editor/blob/main/src/react-utils/useSticky.ts#L37) (`refPageOffset <= stickyOffset`) fails to trigger, causing the sticky header functionality to stop working as expected.

#### Possible Solutions

1. **Utilizing Intersection Observer API**:  
   We can explore using the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to detect when elements enter or exit the viewport. This method would involve observing multiple elements, including those that might dynamically change. However, elements with `position: sticky` may not trigger the observer effectively due to their inherent visibility within the viewport.

2. **Direct Position Comparison**:  
   A simpler approach might involve directly comparing the initial top position of the `ref` component with the `window` scroll position or the scroll position of the nearest parent which has `overflow: auto` or window.

#### Key Considerations
- **Handling Dynamic DOM Changes**: It's crucial to account for elements that might be added to the DOM at runtime, as these could disrupt the layout and the sticky behavior.
- **Flexible Offset Management**: Introducing a CSS variable for the sticky offset allows for more fine-tuned control across different layouts and components, enhancing maintainability and adaptability in varied environments.

makhnatkin:
We should also check the behavior in the Popup

PR #369, opened 2024-09-16
fix(react-utils): fixed useSticky

makhnatkin:
Fixes https://github.com/gravity-ui/markdown-editor/issues/368


src/react-utils/useSticky.ts
```diff
@@ -1,45 +1,139 @@
...
+ *
+ * @param {HTMLElement | null} element - The starting element to search for the CSS variable.
+ * @param {string} variableName - The name of the CSS variable to search for.
+ * @returns {number} - The value of the CSS variable or 0 if not found.
+ */
+const findCssVariableValue = (element: HTMLElement | null, variableName: string): number => {
+    let currentElement = element;
+    while (currentElement) {
+        const value = getComputedStyle(currentElement).getPropertyValue(variableName);
+        if (value) {
```
makhnatkin, line 40: If the value is `0`, then the check will fail.

src/react-utils/useSticky.ts
```diff
@@ -1,45 +1,139 @@
...
+            return parseFloat(value);
+        }
+        currentElement = currentElement.parentElement;
+    }
+    // Fallback to global :root if not found locally
+    return (
+        parseFloat(getComputedStyle(document.documentElement).getPropertyValue(variableName)) || 0
+    );
```
d3m1d0v, line 48: It's can be shorter, like this:

```
let currentElement = element || document.documentElement;
const value = getComputedStyle(currentElement).getPropertyValue(variableName);
return value === '' ? 0 : parseFloat(value); 
```

Because `getComputedStyle().getPropertyValue()` returns closest variable value in DOM tree

src/react-utils/useSticky.ts
```diff
@@ -1,45 +1,139 @@
...
-    'touchend',
-    'pageshow',
-    'load',
-    'orientationchange',
-]);
-
-export function useSticky<T extends HTMLElement>(elemRef: RefObject<T>) {
+import {RefObject, useEffect, useState} from 'react';
+
+import throttle from 'lodash/throttle';
```
d3m1d0v, line 3: plz, use `throttle` from `src/lodash.ts`

src/react-utils/useSticky.ts
```diff
@@ -1,45 +1,139 @@
...
+ * @param {HTMLElement | null} element - The starting element from which to begin the search.
+ * @returns {HTMLElement | Window} - The first parent with overflow: auto/scroll, or window.
+ */
+const findScrollContainer = (element: HTMLElement | null): HTMLElement | Window => {
+    let currentElement = element;
+    while (currentElement) {
+        const overflow = window.getComputedStyle(currentElement).overflow;
+        if (overflow === 'auto' || overflow === 'scroll') {
+            return currentElement;
+        }
```
d3m1d0v, line 22: or `overflow-y: auto | scroll` ?

src/react-utils/useSticky.ts
```diff
@@ -1,45 +1,139 @@
...
+ * @param {HTMLElement | null} element - The starting element to search for the CSS variable.
+ * @param {string} variableName - The name of the CSS variable to search for.
+ * @returns {number} - The value of the CSS variable or 0 if not found.
+ */
+const findCssVariableValue = (element: HTMLElement | null, variableName: string): number => {
+    let currentElement = element;
+    while (currentElement) {
+        const value = getComputedStyle(currentElement).getPropertyValue(variableName);
+        if (value) {
+            return parseFloat(value);
```
d3m1d0v, line 41: If you need to calculate numeric value of css-var, you need to use something like [this function](https://github.com/gravity-ui/markdown-editor/blob/bd85a1c4c073c665c7fa9b8c7277a8441c6f378b/src/bundle/Editor.ts#L538) because is css-vars can store a css expression like `5px + 7px`, so parseFloat or parseInt don't work correctly

PR #412, opened 2024-10-09
fix(Lists): check that list items are being serialized with correct markup

d3m1d0v:
When switch ordered_list –> bullet_list (or vice versa), the list items are serialized with incorrect markup.

This change fix it.

### DIFF
--- a/src/extensions/markdown/Lists/ListsSpecs/serializer.ts
+++ b/src/extensions/markdown/Lists/ListsSpecs/serializer.ts
@@ -2,7 +2,7 @@ import type {Node} from 'prosemirror-model';
 
 import type {SerializerNodeToken} from '../../../../core';
 
-import {ListNode, ListsAttr} from './const';
+import {ListNode, ListsAttr, Markup} from './const';
 
 export const serializerTokens: Record<ListNode, SerializerNodeToken> = {
     [ListNode.ListItem]: (state, node) => {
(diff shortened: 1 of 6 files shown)


makhnatkin (approved):

src/extensions/markdown/Lists/ListsSpecs/serializer.ts
```diff
@@ -2,29 +2,44 @@ import type {Node} from 'prosemirror-model';
...
+    list,
+    type,
+}: {
+    item: Node;
+    list: Node;
+    type: keyof typeof Markup;
+}): string {
+    const defs = Markup[type];
+    let value = item.attrs[ListsAttr.Markup];
+    if (!defs.values.includes(value)) value = list.attrs[ListsAttr.Markup];
```
shevnv, line 42: Looks complicated when you change value of `value` variable and have two equal `if` negative conditions.

I think, positive conditions without reassign are easier to read
```
if (defs.values.includes(item.attrs[ListsAttr.Markup])) {
    return item.attrs[ListsAttr.Markup];
}
if (defs.values.includes(list.attrs[ListsAttr.Markup]) {
    return list.attrs[ListsAttr.Markup];
}
// or maybe you can do same with .find
// [item.attrs[ListsAttr.Markup], list.attrs[ListsAttr.Markup]].find(item => defs.values.includes(item)) || defs.default

return defs.default;
```

PR #416, opened 2024-10-10
feat: Update editor settings

ReFFaT:


src/bundle/settings/index.tsx
```diff
@@ -34,10 +34,12 @@ const bContent = cn('settings-content');
 
 export type EditorSettingsProps = Omit<SettingsContentProps, 'onClose'> & {
     renderPreviewButton?: boolean;
+    settingsVisible?: boolean;
 };
 
 export const EditorSettings = React.memo<EditorSettingsProps>(function EditorSettings(props) {
-    const {className, onShowPreviewChange, showPreview, renderPreviewButton} = props;
+    const {className, onShowPreviewChange, showPreview, renderPreviewButton, settingsVisible} =
```
makhnatkin, line 41: lets add `settingsVisible = true` for compatibility 
```
const {className, onShowPreviewChange, showPreview, renderPreviewButton, settingsVisible = true} =
```

src/bundle/MarkdownEditorView.tsx
```diff
@@ -295,6 +297,9 @@ const MarkupSearchAnchor: React.FC<MarkupSearchAnchorProps> = ({mode}) => (
 function Settings(props: EditorSettingsProps & {stickyToolbar: boolean}) {
     const wrapperRef = useRef<HTMLDivElement>(null);
     const isSticky = useSticky(wrapperRef) && props.toolbarVisibility && props.stickyToolbar;
+
+    if (!props.renderPreviewButton && !props.settingsVisible) return null;
```
makhnatkin, line 301: This looks like breaking changes, I suggest passing the default values so that it works for previous cases when renderPreviewButton and settingsVisible are not set

d3m1d0v:
What problem does this change solve?

ReFFaT:
> What problem does this change solve?

Solves the problem of the preview mode disappearing if you disable the settings in the markup

src/bundle/MarkdownEditorView.tsx
ReFFaT: I replaced the highlighted code fragment, tested it with the main branch (behavior with preview and settings turned off, as well as without props), the behavior was expected

makhnatkin (approved):

PR #448, opened 2024-10-29
perf: disabled yfmLang temporarily for the test

makhnatkin:


makhnatkin:
fixed here https://discuss.codemirror.net/t/the-editor-freezes-when-using-markdown-and-large-text/8752

PR #451, opened 2024-10-31
feat(ImgSize, markup): rework image styling to get more control 

St1ggy:


src/markup/codemirror/files-upload-plugin/plugin.ts
```diff
@@ -148,19 +148,25 @@ class FileUploadPresenter {
...
+        let markup = `![${fileName}](${res.url}`;
+
         if (isImageFile(this.file)) {
             if (this.needDimmensionsForImages) {
                 try {
-                    let {height} = await getImageDimensions(this.file);
-                    height = Math.min(height, IMG_MAX_HEIGHT);
-                    markup = `![${fileName}](${res.url} =x${height})`;
+                    const fileSize = await getImageDimensions(this.file);
+                    const {width, height} = getProportionalSize({
```
makhnatkin, line 157: Let's add an option in the [experiment config](https://github.com/gravity-ui/markdown-editor/blob/main/src/bundle/types.ts#L51) and use it both in WYSIWYG and in markup mode

src/markup/codemirror/files-upload-plugin/plugin.ts
```diff
@@ -148,19 +148,25 @@ class FileUploadPresenter {
...
+                        width: fileSize.width,
+                        height: fileSize.height,
+                        imgMaxHeight: IMG_MAX_HEIGHT,
+                    });
+
+                    markup += ` =${width}x${height}`;
                 } catch (err) {
-                    markup = `![${fileName}](${res.url})`;
+                    console.error(err);
```
makhnatkin, line 174: @d3m1d0v note: I suggest we think about a common logging module

makhnatkin (approved):

Issue #387, opened 2024-09-27
Escape `_` during serialization of wysiwyg content


d3m1d0v:
### Description

Wysiwyg serializer does not escape `_` symbol in content. This may cause differences in the display in wysiwyg and in the html rendered in makrdown-it. For example in case "\_em_": when writing in wysiwyg, this content will have underscores, but after changing the mode to markup and back, in wisywig it would be "_em_". Markdown-it parse it as italic emphasis.

### Proposal

Need to add the underscore character `_` to the list for escaping in serializer to markdown

PR #452, opened 2024-10-31
fix(core): added escaping the underscore character via serializing

u4aew:
Add escape characters "_"

issue [387](https://github.com/gravity-ui/markdown-editor/issues/387)


src/core/markdown/MarkdownSerializer.js
```diff
@@ -301,7 +301,7 @@ export class MarkdownSerializerState {
     // content. If `startOfLine` is true, also escape characters that
     // have special meaning only at the start of the line.
     esc(str, startOfLine) {
-        const escRegexp = this.options?.commonEscape || /[`\^+*\\\|~\[\]\{\}<>\$]/g;
+        const escRegexp = this.options?.commonEscape || /[`\^+*\\\|~\[\]\{\}<>\$_]/g;
```
makhnatkin, line 304: Let's add tests for this case to `markdown-editor/src/core/markdown/Markdown.test.ts`. What do you think?

u4aew: Yes, okay, I'll add tests

makhnatkin (approved):

makhnatkin:
@u4aew thx!

PR #456, opened 2024-11-02
fix(markup): correct key formatting for codemirror keymap

d3m1d0v:
In markup mode Escape key handler was not called, so `cancel` event was never emitted


St1ggy (approved):

makhnatkin:
> In markup mode Escape key handler was not called, so `cancel` event was never emitted

Note: In Codemirror 5, the shortcut format was different from that used in Prosemirror. After switching to Codemiror version 6, this was no longer necessary.

PR #464, opened 2024-11-08
feat(Image): convert pasted image urls to images

kseniya57:
Added the functionality for converting pasted urls to images

### DIFF
--- a/src/extensions/markdown/Image/imageUrlPaste/index.ts
+++ b/src/extensions/markdown/Image/imageUrlPaste/index.ts
@@ -1,0 +1,54 @@
+import {Plugin} from 'prosemirror-state';
+
+import {ExtensionAuto} from '../../../../core';
+import {DataTransferType} from '../../../behavior/Clipboard/utils';
+import {imageType} from '../ImageSpecs';
+
+export type ImageUrlPasteOptions = {
+    /**
+     * The function, used to determine if the pasted text is the image url and should be inserted as an image
+     */
+    parseInsertedUrlAsImage?: (text: string) => {imageUrl: string; title?: string} | null;
+};
+
+export const imageUrlPaste: ExtensionAuto<ImageUrlPasteOptions> = (builder, opts) => {
+    builder.addPlugin(
+        () =>
+            new Plugin({
+                props: {
+                    handleDOMEvents: {
+                        paste(view, e) {
+                            if (
+                                !opts.parseInsertedUrlAsImage ||
+                                !e.clipboardData ||
+                                view.state.selection.$from.parent.type.spec.code
+                            )
+                                return false;
+
+                            const {imageUrl, title} =
+                                opts.parseInsertedUrlAsImage(
+                                    e.clipboardData.getData(DataTransferType.Text) ?? '',
+                                ) || {};
+
+                            if (!imageUrl) {
+                                return false;
+                            }
+
+                            e.preventDefault();
+
+                            const imageNode = imageType(view.state.schema).create({
+                                src: imageUrl,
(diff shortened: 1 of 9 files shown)


src/extensions/yfm/ImgSize/const.ts
```diff
@@ -6,3 +6,13 @@ export {imageNodeName, addImageAction} from '../../markdown/Image/const';
 export const IMG_MAX_HEIGHT = 600; //px
 export type ImageRendererState = {linkAdded: boolean};
 export const imageRendererKey = new PluginKey<ImageRendererState>('imageRenderer');
+
+const knownImageHostsRegexString = '(jing|avatars\\.mds)';
```
makhnatkin, line 10: I suggest putting the regular expression into the parameters of the extension rather than hardcoding it into the source code of the markdown editor.

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -39,6 +41,28 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
+                                    alt: imageUrl.match(imageNameRegex)?.[1],
+                                });
+
+                                view.dispatch(
+                                    view.state.tr.insert(view.state.tr.selection.from, [
+                                        view.state.schema.text(' '),
+                                        imageNode,
+                                        view.state.schema.text(' '),
+                                    ]),
+                                );
```
makhnatkin, line 65: `return true;` ?
In theory, we would expect to return true, right?

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -39,6 +41,28 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
+
+                                e.preventDefault();
+
+                                const imageNode = imageType(view.state.schema).create({
+                                    src: imageUrl,
+                                    alt: imageUrl.match(imageNameRegex)?.[1],
+                                });
+
+                                view.dispatch(
+                                    view.state.tr.insert(view.state.tr.selection.from, [
```
makhnatkin, line 60: Why is a `view.state.tr.insert` and a `view.state.schema.text(' ')` used here, ideally the code should be reduced to one view, as [here](https://github.com/gravity-ui/markdown-editor/pull/464/files#diff-550af0eac60195e2f489e0ca052ef38313c36973ddcf71a2adde42412e9ee4bdR24)

makhnatkin:
The current solution is suitable. At the same time, we plan to implement another approach in the future: when a link is inserted, a popup will appear and the user will be asked to select the type of insertion (similar to Notion).

src/extensions/markdown/Image/imageUrlPaste/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type ImageUrlPasteOptions = {
+    /**
+     * The function, used to determine if the pasted text is the image url and should be inserted as an image
+     */
+    shouldInsertPastedUrlAsImage?: (text: string) => boolean;
+    /**
+     * The function, used to get the image alternative text from the pasted image url
+     */
+    getImageNameFromPastedUrl?: (imageUrl: string) => string;
```
makhnatkin, line 15: I suggest combining these two functions into one that returns either an object or `null`, the object may contains the title

for example:
```
parseInsertedUrlAsImage?: (text: string) => {imageUrl: string, title?: string} | null;
```

makhnatkin (approved):

PR #466, opened 2024-11-08
fix(Italic): fix italic toMd symbols

St1ggy:


St1ggy:
image
image
image
image

WYSIWYG and MD now have different symbols

PR #469, opened 2024-11-12
feat(markup): convert pasted urls to images

kseniya57:


src/bundle/Editor.ts
```diff
@@ -313,6 +319,8 @@ export class EditorImpl extends SafeEventEmitter<EventMapInt> implements EditorI
 
         this.#renderStorage = opts.renderStorage;
         this.#fileUploadHandler = handlers.uploadFile;
+        this.#parseInsertedUrlAsImage =
+            wysiwygConfig.extensionOptions?.imgSize?.parseInsertedUrlAsImage;
```
makhnatkin, line 323: I suggest creating the same prop in [markup config](https://github.com/gravity-ui/markdown-editor/blob/main/src/bundle/types.ts#L68), so as not to take it from WYSIWYG config for for markup.

src/markup/codemirror/create.ts
```diff
@@ -145,6 +149,28 @@ export function createCodemirror(params: CreateCodemirrorParams) {
...
+                            event.clipboardData.getData(DataTransferType.Text) ?? '',
+                        ) || {};
+
+                    if (!imageUrl) {
+                        return;
+                    }
+
+                    event.preventDefault();
+
+                    insertImages([
```
makhnatkin, line 165: Note: As far as I understand, when inserting this image, its size may exceed the size of the container, which may cause empty areas to appear on the screen. As a result, users have two different experiences: when inserting a file normally, the image fits into the specified dimensions, and when inserting a link from above, there may be empty space. Perhaps in the future it is worth preloading the image to correctly determine its size.

src/bundle/types.ts
```diff
@@ -98,6 +98,10 @@ export type MarkdownEditorMarkupConfig = {
     languageData?: YfmLangOptions['languageData'];
     /** Config for @codemirror/autocomplete https://codemirror.net/docs/ref/#autocomplete.autocompletion%5Econfig */
     autocompletion?: CreateCodemirrorParams['autocompletion'];
+    /**
+     * The function, used to determine if the pasted text is the image url and should be inserted as an image
+     */
+    parseInsertedUrlAsImage?: (text: string) => {imageUrl: string; title?: string} | null;
```
makhnatkin, line 104: nit:
Lets create type for `(text: string) => {imageUrl: string; title?: string} | null`
and we will use it in WYSIWYG and markup

makhnatkin (approved):

makhnatkin (approved):

PR #483, opened 2024-11-18
fix(bundle): added emoji to toolbar actions for full preset in markup mode

makhnatkin:


src/bundle/config/markup.tsx
```diff
@@ -59,8 +59,9 @@ import {ActionName} from './action-names';
 import {icons} from './icons';
 
 const noop = () => {};
-const isActiveFn = () => false;
-const isEnableFn = () => true;
+const inactive = () => false;
+const enable = () => true;
+const disbale = () => false;
```
d3m1d0v, line 64: dis**a**ble

src/bundle/config/wysiwyg.ts
```diff
@@ -531,6 +541,7 @@ export const wToolbarConfig: WToolbarData = [
         },
     ],
     [wImageItemData, wFileItemData, wTableItemData, wCheckboxItemData],
+    [wMermaidItemData, wYfmHtmlBlockItemData],
```
d3m1d0v, line 544: mermaid is not enabled by default

d3m1d0v (approved):

PR #484, opened 2024-11-19
refactor: updated demo structure

makhnatkin:


d3m1d0v (approved):
Plz, change name of PR – it should starts with `chore:` or `chore(demo):`.
Because prefix `refactor:` refers to source code refactoring, but demo and playground are not part of source code.

PR #485, opened 2024-11-19
feat(core): smartindent on paste

obenjiro:
Inserting text with multiple paragraphs into the citation block in markdown mode

Now in markdown mode, you can copy and paste something quite complex without breaking MD syntax
(lots of paragraphs of text, quotes, lists, tabs, etc.)

https://github.com/user-attachments/assets/ef51adfb-f929-4199-bf38-efba1956d2d1

https://github.com/user-attachments/assets/6b84b5e5-2205-4f9b-9f89-31921f172de5


src/markup/codemirror/paste-handler/utils.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Parses markdown-style markers from the start of a line
+ * Returns an array of markers found:
+ * - '  ' for indentation
+ * - '> ' for blockquotes
+ * - '* ' or '- ' for list items
```
d3m1d0v, line 6: list items marks with `*` or `-` or `+`
so you need to add `+` to this list

d3m1d0v: and what about ordered lists?

src/markup/codemirror/create.ts
```diff
@@ -152,7 +153,18 @@ export function createCodemirror(params: CreateCodemirrorParams) {
...
-                if (event.clipboardData && parseInsertedUrlAsImage) {
+                if (!event.clipboardData) return false;
+
+                const pastedText = event.clipboardData.getData('text/plain');
+
+                // Handle markdown-aware pasting
+                if (handleMarkdownPaste(pastedText, editor)) {
+                    event.preventDefault();
+                    return true;
+                }
```
d3m1d0v, line 164: I'm guessing `handleMarkdownPaste()` should be after `parseInsertedUrlAsImage()`

src/markup/codemirror/create.ts
```diff
@@ -171,7 +183,11 @@ export function createCodemirror(params: CreateCodemirrorParams) {
                             title,
                         },
                     ])(editor);
+
+                    return true;
                 }
+
+                return true;
```
d3m1d0v, line 190: We want to keep default paste handling by codemirror if we haven't done anything ourselves. So, should be `return false`

d3m1d0v:
@obenjiro 

Good job! It looks really good 👍

PR #492, opened 2024-11-21
feat(core): markdown smart wrapInLine

obenjiro:
Third an last one in the series of PR's to markdown. This one is also fixing consistency between WYSIWYG and MARKDOWN modes. Now we can use buttons like checkbox and quote while maintaining integrity of markdown syntax in hardcore examples.

(tabs, lists, etc.)

https://github.com/user-attachments/assets/aace31ee-a9a3-49fb-89f2-37f75b5f6aa7

PS: I will add test a little bit later

### DIFF
--- a/src/markup/commands/helpers.ts
+++ b/src/markup/commands/helpers.ts
@@ -209,12 +209,86 @@ export const wrapPerLine =
 
             const isSingleEmptyLine =
                 range.anchor === range.head && state.doc.lineAt(range.head).length === 0;
+
             if (isSingleEmptyLine) {
-                changes.push({from: range.head, insert: before});
+                // Get markers from previous line for empty line case
+                if (range.head > 1) {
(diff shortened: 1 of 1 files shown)


src/markup/commands/helpers.ts
```diff
@@ -209,12 +209,86 @@ export const wrapPerLine =
...
+                    const markers = parseMarkers(prevLine.text);
+                    const indentation = markers.join('');
+                    changes.push({from: range.head + indentation.length, insert: before});
+                } else {
+                    changes.push({from: range.head, insert: before});
+                }
             } else {
+                // Calculate indentation from the first line
+                const firstLine = state.doc.lineAt(range.from);
+                let indentation = '';
```
makhnatkin, line 226: Let`s move code to `getIndentation ` or `calculacteIndentation ` or `parseIndentation`
and tests plz

PR #501, opened 2024-11-27
Pamiant: add empty row

PMAWorks:


makhnatkin (changes requested):
I propose to change the pr according to the [recommendations](https://github.com/gravity-ui/markdown-editor/blob/main/docs/guidelines-contributions.md)

PR #506, opened 2024-11-29
feat (core): add empty row placeholder

PMAWorks:
Adds a flag "emptyRowPlaceholder" where a placeholder will be displayed in every empty line that is in focus, not just in an empty document

BEFORE
image

AFTER
image

### DIFF
--- a/demo/components/Playground.tsx
+++ b/demo/components/Playground.tsx
@@ -79,6 +79,7 @@ export type PlaygroundProps = {
     breaks?: boolean;
     linkify?: boolean;
     linkifyTlds?: string | string[];
+    emptyRowPlaceholder?: boolean;
     sanitizeHtml?: boolean;
     prepareRawMarkup?: boolean;
     splitModeOrientation?: 'horizontal' | 'vertical' | false;
@@ -124,6 +125,7 @@ export const Playground = React.memo<PlaygroundProps>((props) => {
         initialSplitModeEnabled,
         settingsVisible,
         allowHTML,
+        emptyRowPlaceholder,
         breaks,
         linkify,
         linkifyTlds,
@@ -185,6 +187,9 @@ export const Playground = React.memo<PlaygroundProps>((props) => {
             needToSetDimensionsForUploadedImages,
             renderPreview: renderPreviewDefined ? renderPreview : undefined,
             fileUploadHandler,
+            md: {
+                emptyRowPlaceholder: emptyRowPlaceholder,
+            },
             experimental: {
                 ...experimental,
                 directiveSyntax,
--- a/src/bundle/types.ts
+++ b/src/bundle/types.ts
@@ -28,6 +28,7 @@ export type ParseInsertedUrlAsImage = (text: string) => {imageUrl: string; title
 
 export type MarkdownEditorMdOptions = {
     html?: boolean;
+    emptyRowPlaceholder?: boolean;
     breaks?: boolean;
     linkify?: boolean;
     linkifyTlds?: string | string[];
--- a/src/bundle/wysiwyg-preset.ts
+++ b/src/bundle/wysiwyg-preset.ts
@@ -27,6 +27,7 @@ export type BundlePresetOptions = ExtensionsOptions &
         preset: MarkdownEditorPreset;
         mdBreaks?: boolean;
         fileUploadHandler?: FileUploadHandler;
+        emptyRowPlaceholder?: boolean;
         /**
          * If we need to set dimensions for uploaded images
          *
@@ -64,7 +65,10 @@ export const BundlePreset: ExtensionAuto<BundlePresetOptions> = (builder, opts) 
             paragraphKey: f.toPM(A.Text),
             paragraphPlaceholder: (node: Node, parent?: Node | null) => {
                 const isDocEmpty =
-                    !node.text && parent?.type.name === BaseNode.Doc && parent.childCount === 1;
+                    !node.text &&
+                    ((opts.emptyRowPlaceholder && parent?.type.name === 'doc') ||
+                        (parent?.type.name === BaseNode.Doc && parent.childCount === 1));
+
                 return isDocEmpty ? i18nPlaceholder('doc_empty') : null;
             },
             ...opts.baseSchema,
--- a/src/bundle/useMarkdownEditor.ts
+++ b/src/bundle/useMarkdownEditor.ts
@@ -59,6 +59,7 @@ export function useMarkdownEditor<T extends object = {}>(
                     editor.emit('submit', null);
                     return true;
                 },
+                emptyRowPlaceholder: md.emptyRowPlaceholder,
                 mdBreaks: breaks,
                 fileUploadHandler: uploadFile,
                 needToSetDimensionsForUploadedImages,


demo/components/Playground.tsx
```diff
@@ -185,6 +187,9 @@ export const Playground = React.memo<PlaygroundProps>((props) => {
             needToSetDimensionsForUploadedImages,
             renderPreview: renderPreviewDefined ? renderPreview : undefined,
             fileUploadHandler,
+            md: {
+                emptyRowPlaceholder: emptyRowPlaceholder,
+            },
```
d3m1d0v, line 192: I propose some changes:
1. move setting from `md` to `wysiwygConfig`
2. in `wysiwygConfig` add new optional property for placeholder configuration, that have structure like that:
  ```
  value: string | () => string;
  behavior: 'empty-doc' | 'empty-row'
  ```

src/bundle/wysiwyg-preset.ts
```diff
@@ -64,7 +65,10 @@ export const BundlePreset: ExtensionAuto<BundlePresetOptions> = (builder, opts)
             paragraphKey: f.toPM(A.Text),
             paragraphPlaceholder: (node: Node, parent?: Node | null) => {
                 const isDocEmpty =
-                    !node.text && parent?.type.name === BaseNode.Doc && parent.childCount === 1;
+                    !node.text &&
+                    ((opts.emptyRowPlaceholder && parent?.type.name === 'doc') ||
```
makhnatkin, line 69: `opts.emptyRowPlaceholder ` 
It seems that the logic of the behavior of an empty line for a paragraph and for a dock should behave differently, I suggest creating 2 variables

makhnatkin (changes requested):

src/bundle/types.ts
```diff
@@ -26,6 +26,11 @@ export type RenderPreview = (params: RenderPreviewParams) => ReactNode;
 
 export type ParseInsertedUrlAsImage = (text: string) => {imageUrl: string; title?: string} | null;
 
+export type WysywigPlaceholderOptions = {
```
d3m1d0v, line 29: Wys**i**wyg

src/bundle/types.ts
```diff
@@ -26,6 +26,11 @@ export type RenderPreview = (params: RenderPreviewParams) => ReactNode;
...
+    value?: string | (() => string);
+    behavior: 'empty-doc' | 'empty-row';
```
d3m1d0v, line 31: should `behavior` option also be optional? In my opinion, yes

d3m1d0v: in any case, plz add comment about default value

src/bundle/wysiwyg-preset.ts
```diff
@@ -63,9 +64,26 @@ export const BundlePreset: ExtensionAuto<BundlePresetOptions> = (builder, opts)
         baseSchema: {
             paragraphKey: f.toPM(A.Text),
             paragraphPlaceholder: (node: Node, parent?: Node | null) => {
-                const isDocEmpty =
-                    !node.text && parent?.type.name === BaseNode.Doc && parent.childCount === 1;
-                return isDocEmpty ? i18nPlaceholder('doc_empty') : null;
+                let isPlaceholderShown = false;
+
+                if (opts.placeholderOptions?.behavior === 'empty-row') {
+                    isPlaceholderShown = !node.text && parent?.type.name === 'doc';
```
d3m1d0v, line 70: `parent?.type.name === 'doc'` – this means that placeholder will be displayed only if empty paragraph is direct child of doc (is located top level in document tree).

Perhaps when `behavior:'empty-row'` placeholder should be diplayed in any empty paragraph? For example, inside notes, cuts, qoutes and others?

src/bundle/wysiwyg-preset.ts
```diff
@@ -63,9 +64,26 @@ export const BundlePreset: ExtensionAuto<BundlePresetOptions> = (builder, opts)
...
+
+                if (opts.placeholderOptions?.value) {
+                    if (typeof opts.placeholderOptions?.value === 'string') {
+                        text = opts.placeholderOptions?.value;
+                    } else {
+                        text = opts.placeholderOptions?.value();
+                    }
+                }
+
+                return isPlaceholderShown ? text : null;
```
d3m1d0v, line 86: Let's rewrite this piece of code – make it more readable and a little more optimal

```js
const {behavior, value} = opts.placeholderOptions || {};

const isRowEmpty = !node.text && parent?.type.name === BaseNode.Doc;
const isDocEmpty = isRowEmpty && parent.childCount === 1;

const showPlaceholder = behavior === 'empty-row' ? isRowEmpty : isDocEmpty;

if (!showPlaceholder) return null;

return typeof value === 'function' ? value() : value ?? i18nPlaceholder('doc_empty');
```

src/bundle/wysiwyg-preset.ts
PMAWorks: The idea is that we should only consider an empty string to be present if it is actually part of the document. We do not want to display placeholders inside blocks

src/bundle/types.ts
```diff
@@ -26,6 +26,12 @@ export type RenderPreview = (params: RenderPreviewParams) => ReactNode;
 
 export type ParseInsertedUrlAsImage = (text: string) => {imageUrl: string; title?: string} | null;
 
+export type WysiwygPlaceholderOptions = {
+    value?: string | (() => string);
+    /** Default – empty-doc */
+    behavior?: 'empty-doc' | 'empty-doc-row' | 'empty-row';
```
makhnatkin, line 32: Lets add something like that
https://github.com/gravity-ui/markdown-editor/blob/main/src/bundle/types.ts#L86

```
empty-doc
empty-row
empty-row-top-level
```

src/bundle/wysiwyg-preset.ts
```diff
@@ -63,9 +64,22 @@ export const BundlePreset: ExtensionAuto<BundlePresetOptions> = (builder, opts)
         baseSchema: {
             paragraphKey: f.toPM(A.Text),
             paragraphPlaceholder: (node: Node, parent?: Node | null) => {
-                const isDocEmpty =
-                    !node.text && parent?.type.name === BaseNode.Doc && parent.childCount === 1;
-                return isDocEmpty ? i18nPlaceholder('doc_empty') : null;
+                const {value, behavior} = opts.placeholderOptions || {};
+
+                const emptyList = {
```
makhnatkin, line 69: `empties`
`emptyValues`
`emptyEntries`

makhnatkin (approved):

src/bundle/types.ts
```diff
@@ -26,6 +26,17 @@ export type RenderPreview = (params: RenderPreviewParams) => ReactNode;
 
 export type ParseInsertedUrlAsImage = (text: string) => {imageUrl: string; title?: string} | null;
 
+export type WysiwygPlaceholderOptions = {
+    value?: string | (() => string);
+    /** Default – empty-doc 
+    Values:
+    - 'empty-doc' – the placeholder will only be shown in an empty document;
+    - 'empty-row-top-level' – the placeholder will be displayed in an empty line, which is located at the top level of the document;
+    - 'empty-row' – Заполнитель будет отображаться в любой пустой строке документа;
```
makhnatkin, line 35: @PMAWorks please use English for comments

makhnatkin: maybe like this:
```
- 'empty-doc' – The placeholder will only be shown when the document is completely empty.
- 'empty-row-top-level' – The placeholder will be displayed in an empty line that is at the top level of the document structure.
- 'empty-row' – The placeholder will be shown in any empty line within the document, regardless of its nesting level.
```

makhnatkin (approved):

makhnatkin (approved):

PR #530, opened 2024-12-19
feat(markup): smart re-indent on paste

obenjiro:
Inserting text with multiple paragraphs into the citation block in markdown mode

Now in markdown mode, you can copy and paste something quite complex without breaking MD syntax
(lots of paragraphs of text, quotes, lists, tabs, etc.)

This is continuation of https://github.com/gravity-ui/markdown-editor/pull/485


src/markup/codemirror/paste-handler/index.ts
d3m1d0v, line 1: now this is not a paste handler. Plz, rename the folder to more appropriate name

d3m1d0v (approved):

src/markup/codemirror/paste-handler/index.ts
obenjiro: fixed

PR #533, opened 2024-12-20
feat(Selection): made all top-level nodes selectable

whiteformed:
Make all top-level nodes selectable


src/extensions/behavior/Selection/selection.ts
```diff
@@ -99,18 +99,16 @@ const getTopLevelNodesFromSelection = (selection: Selection, doc: Node) => {
...
             const withinSelection = from <= pos && pos + node.nodeSize <= to;
-            if (
-                node &&
-                node.type.name !== 'paragraph' &&
-                !node.isText &&
-                node.type.spec.selectable &&
-                withinSelection
-            ) {
+
+            if (node && !node.isText && node.type.spec.selectable && withinSelection) {
```
d3m1d0v, line 106: Default value of `spec.selectable` is `true`
image

You should replace check of `node.type.spec.selectable` with `node.type.spec.selectable !== false`

d3m1d0v (approved):

PR #538, opened 2024-12-25
fix(bundle): fixed imports with relative paths

whiteformed:
Fix relative import @ Editor

### DIFF
--- a/src/bundle/Editor.ts
+++ b/src/bundle/Editor.ts
@@ -4,9 +4,6 @@ import {EditorView as CMEditorView} from '@codemirror/view';
 import {TextSelection} from 'prosemirror-state';
 import {EditorView as PMEditorView} from 'prosemirror-view';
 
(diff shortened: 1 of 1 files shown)


src/bundle/Editor.ts
```diff
@@ -4,9 +4,9 @@ import {EditorView as CMEditorView} from '@codemirror/view';
 import {TextSelection} from 'prosemirror-state';
 import {EditorView as PMEditorView} from 'prosemirror-view';
 
-import {TransformFn} from 'src/core/markdown/ProseMirrorTransformer';
-
+import {TransformFn} from '../../src/core/markdown/ProseMirrorTransformer';
 import {getAutocompleteConfig} from '../../src/markup/codemirror/autocomplete';
+
```
d3m1d0v, line 9: Change relative paths to `../core/...` and `../markup/...`

d3m1d0v (changes requested):

d3m1d0v (approved):

PR #544, opened 2024-12-27
fix(toolbar): use helper to adjust z-index for popups (issue #298)

### LINKED ISSUE #298: Help tooltip is hidden under the toolbar on small screens
On small screens help tooltip with Markdown markup is hidden under the toolbar.

image

**Steps to Reproduce:**

- Open the Editor on small screen (height <= 400px).
- Scroll down editor content to switch toolbar into floating mode
- Hover over the settings icon in the top right corner.
- Hover over the help icon near the Markdown markup text - top of the tooltip will be hidden under the toolbar

**Expected Behavior:**
Tooltip will be on top of the toolbar.

**Actual Behavior:**
Tooltip is under the toolbar.

**Possible Solution:**
Play with [z-index](https://github.com/gravity-ui/markdown-editor/blob/bfeb1380caec8087c0fd0148a23a0eb4197510c9/src/styles/_zindex.scss#L6) :)

dpolevodin:
fixes #298 
z-index for the sticky-toolbar has been lowered from 2000 to 990, since the HelpMark tooltip index is set to 1000 in uikit
Added z-index system and updated readme.
Changed current z-index values in project by z-index mixin

image


README.md
```diff
@@ -82,6 +84,32 @@ configure({
...
+### Usage
+
+To apply z-index to components, use the z-index mixin, passing the key as an argument:
+
+```scss
+@use 'styles/mixins.scss';
+
+.tooltip {
+  @include mixins.z-index('forefront');
+}
```
makhnatkin, line 110: add to docs plz too

src/styles/_zindex.scss
```diff
@@ -1,6 +1,10 @@
...
-$sticky-toolbar: 2000; // more than has popup
+// z-index system values
+$z-index-map: (
+    'background': -1,
+    'default': 0,
+    'forefront': 1,
+    'img-settings-button': 2,
+    'table-view-button': 100,
+    'table-cell-button': 110,
+    'sticky-toolbar': 990,
```
makhnatkin, line 9: 1. header sticky toolbar+
2. header sticky toolbar
3. context toolbar+ (popups, hints)
4. context toolbar
5. suggest popup+ (for hints), (`...` buttons menu)
6. suggest popup
7. block panel popup

context is popup from text selection
suggest is `/`, `:`

makhnatkin: if you need to change z-index in uikit component use style https://github.com/gravity-ui/uikit/blob/main/src/components/types.ts#L4C5-L4C27

like this:
```
style={{
   // ...otherStyles,
   zIndex: var(--md-editor-z-index-level-suggest);
}}
```

dpolevodin:
> @dpolevodin could you please outline the steps you believe we should take to solve the task at this stage?
@makhnatkin Hi! I would split the task into two parts:
1. Solving the bug when a popup with hints from HelpMark does not overlap the toolbar in the sticky state.
I see the solution to this problem in adding a zIndex prop to the Popover component, since HelpMark uses popoverProps. But for some reason, popoverProps did not include the zIndex property when inheriting properties from PopupProps. I updated the PR using the required zIndex variable for HelpMark inside popoverProps. In fact, this solution works and the required zIndex is applied to the popover from HelpMark, but it causes a TypeScript error: TS2353: Object literal may only specify known properties, and zIndex does not exist in type Omit<PopoverProps, "children"> precisely because the properties were not added to the ui-kit Popover component. I am attaching a screenshot
image

UI:
image

dpolevodin:
> > @dpolevodin could you please outline the steps you believe we should take to solve the task at this stage?
> > @makhnatkin Hi! I would split the task into two parts:
> 
> 1. Solving the bug when a popup with hints from HelpMark does not overlap the toolbar in the sticky state.
>    I see the solution to this problem in adding a zIndex prop to the Popover component, since HelpMark uses popoverProps. But for some reason, popoverProps did not include the zIndex property when inheriting properties from PopupProps. I updated the PR using the required zIndex variable for HelpMark inside popoverProps. In fact, this solution works and the required zIndex is applied to the popover from HelpMark, but it causes a TypeScript error: TS2353: Object literal may only specify known properties, and zIndex does not exist in type Omit<PopoverProps, "children"> precisely because the properties were not added to the ui-kit Popover component. I am attaching a screenshot
>    image
> 
> Now, it works correctly on UI: image



> @dpolevodin could you please outline the steps you believe we should take to solve the task at this stage?

The second part. It is necessary to work out the solution for layers. In general, I see two options:
1. Leave it approximately as it is now in the PR, perhaps think about more generalized names of variables, but in essence, it will be in the form of storing information about the zIndex value for each layer in one file and further adding the necessary index in the component styles using a mixin, as indicated in the PR in Readme, example:
```scss
.tooltip { 
	@include mixins.z-index('forefront'); 
}
```
2. Another option, which is more complex and "heavy", in my opinion in terms of performance, is to create a context and a hook for managing zIndex inside a react component. It might look something like this
``` // ZIndexContext.tsx
import React, { createContext, useContext, useState } from 'react';

type ZIndexContextType = {
  getZIndex: (layer: string) => number;
  bringToFront: (layer: string) => void;
};

const ZIndexContext = createContext<ZIndexContextType | undefined>(undefined);

export const ZIndexProvider: React.FC<{children: React.ReactNode}> = ({ children }) => {
  const [layers, setLayers] = useState<Record<string, number>>({
    deep: -1,
    default: 1,
    dropdown: 100,
    modal: 500,
    notification: 700,
    tooltip: 800,
    max: 9999
  });

  const bringToFront = (layer: string) => {
    setLayers(prev => {
      const currentMax = Math.max(...Object.values(prev));
      return {
        ...prev,
        [layer]: currentMax + 1
      };
    });
  };

  const getZIndex = (layer: string) => layers[layer] || 1;

  return (
    <ZIndexContext.Provider value={{ getZIndex, bringToFront }}>
      {children}
    </ZIndexContext.Provider>
  );
};

export const useZIndex = () => {
  const context = useContext(ZIndexContext);
  if (!context) {
    throw new Error('useZIndex must be used within a ZIndexProvider');
  }
  return context;
}; ```
And the use in the component will probably look like this, using the example of a modal window:
```
// Modal.tsx
import React, { useEffect } from 'react';
import { useZIndex } from './ZIndexContext';
import './Modal.scss';

export const Modal: React.FC = ({ children }) => {
  const { getZIndex, bringToFront } = useZIndex();
  const zIndex = getZIndex('modal');

  useEffect(() => {
    bringToFront('modal');
  }, []);

  return (
    <div className="modal" style={{ zIndex }}>
      {children}
    </div>
  );
};
```
And you will also need to wrap the application in ZIndexProvider like this:
``` export const App = () => {
  return (
    <ZIndexProvider>
      {/* app */}
      <Modal>Modal example</Modal>
    </ZIndexProvider>
  );
};
```
In my opinion, the second option has a complicated logic, in itself can affect the performance due to calling additional hooks on the UI and setting zIndex via style and, possibly, cause additional problems. Although it may look like a more dynamic and flexible solution.

Perhaps you will have even better suggestions :)

makhnatkin:
@dpolevodin lets collect test cases with screenshots of many open popups/tooltips to determine zIndex values ​​and identify possible problems with incorrect layer order

dpolevodin:
> @dpolevodin lets collect test cases with screenshots of many open popups/tooltips to determine zIndex values ​​and identify possible problems with incorrect layer order

@makhnatkin Hi! yes, this is the next step. I will collect it soon

dpolevodin:
> 3\. Collect test cases with screenshots of many open popups/tooltips to determine zIndex values ​​and identify possible problems with incorrect layer order:

1. header sticky toolbar+
- HelpMark (zIndex: 1000) with text variants
image
- text color popup (zIndex: 1000)
image
-Heading popup (zIndex: 100)
image

dpolevodin:
@makhnatkin 
I synced the branch with the main branch to consistently implement this task from the beginning.

Currently, when the toolbar enters the stick position, all its elements with a popup are hidden under the toolbar itself (see "As is").

This PR adds zindexes for tooltip elements in the popup format, which add values ​​to the toolbar's zindex when it's in the stick position (see "To be")

**"As is"**
image
image

**To be"**
image
image

src/common/layersMap.ts
```diff
@@ -0,0 +1,3 @@
+export const CLASSNAME_BY_LAYER_MAP = {
+    STICKY_TOOLBAR: 'g-md-editor-sticky',
```
makhnatkin: data-layout = 'sticky-toolbar'

src/bundle/settings/index.tsx
```diff
@@ -174,6 +178,10 @@ const SettingsContent: React.FC<SettingsContentProps> = function SettingsContent
                                 popoverProps={{
                                     placement: mdHelpPlacement,
                                     modal: false,
+                                    zIndex: getTargetZIndex(
+                                        CLASSNAME_BY_LAYER_MAP.STICKY_TOOLBAR,
+                                        20,
```
makhnatkin: 20?

src/bundle/ToolbarView.tsx
```diff
@@ -3,6 +3,8 @@ import {useLayoutEffect, useRef} from 'react';
 import type {QAProps} from '@gravity-ui/uikit';
 import {useUpdate} from 'react-use';
 
+import {LAYOUT_MAP} from 'src/common/layoutMap';
```
makhnatkin: @dpolevodin @d3m1d0v maybe `LAYOUT`

makhnatkin (approved):

package-lock.json
```diff
@@ -39,7 +39,7 @@
         "prosemirror-commands": "^1.6.2",
         "prosemirror-dropcursor": "^1.8.1",
         "prosemirror-history": "^1.4.1",
-        "prosemirror-inputrules": "^1.5.0",
+        "prosemirror-inputrules": "1.5.0",
```
makhnatkin: revert package-lock plz

PR #545, opened 2024-12-28
fix(nodes): fixed switching a non-empty cut block in wysiwyg

ReFFaT:
There was a problem that when using an empty block of code in cut and switching to markup mode and back, he does not see this block and puts the placeholder text. The problem can be solved by checking for a paragraph, since this is an empty node by default, now we consider a cut to be empty if there is no content in it and the only node in it is node = "paragraph", that is, if there is no text in the cut, but there is a block of code or something else, then it will not being considered empty also fixes a problem for the note node.

### DIFF
--- a/src/utils/nodes.ts
+++ b/src/utils/nodes.ts
@@ -23,7 +23,9 @@ export function findFirstTextblockChild(
 export const isNodeEmpty = (node: Node) => {
     const emptyChildren = findChildren(
         node,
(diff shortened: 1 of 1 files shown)


makhnatkin:
@ReFFaT add tests plz in next PR

makhnatkin (approved):

d3m1d0v (approved):

PR #573, opened 2025-01-28
refactor(core): converted MarkdownSerializer implementation from JS to TS

makhnatkin:
During the conversion to TypeScript, tests, type checks, and runtime behavior revealed no issues.

1. In the `markString` method, a check was added to determine whether parent is a `Node` or a`Fragment`. `parent.content` is used for nodes, and `parent` itself is used for fragments.
2. In the `renderInline` method, a `node.withText` method not present in the official `Node` interface was used. A comment has been added to clarify this.


makhnatkin:
https://github.com/gravity-ui/markdown-editor/pull/596

PR #576, opened 2025-01-29
feat(actions): add previews

whiteformed:

PR #582, opened 2025-02-05
feat(actions): add heading previews

vladKrk:
Add preview image for heading in toolbar
image

It's next step of https://github.com/gravity-ui/markdown-editor/pull/576


src/toolbar/ToolbarListButton.tsx
```diff
@@ -113,6 +124,10 @@ export function ToolbarListButton<E>({
 
                             const titleText = isFunction(title) ? title() : title;
                             const hintText = isFunction(hint) ? hint() : hint;
+                            const preview = previews[id as keyof typeof ActionName];
```
makhnatkin, line 127: runtime undefined possible  
maybe like:
```
const preview = previews[id as keyof typeof ActionName] ?? defaultPreview;
```

src/toolbar/ToolbarListButton.tsx
```diff
@@ -133,38 +148,50 @@ export function ToolbarListButton<E>({
...
-                                                    <HelpPopover
-                                                        className={b('hint')}
-                                                        content={hintText}
-                                                    />
-                                                )}
+                                        <Menu.Item
+                                            key={id}
+                                            active={isActive(editor)}
+                                            disabled={!isEnable(editor)}
+                                            onClick={() => {
```
makhnatkin, line 164: lets move to handler

```
const handleClick = () => {
  hide();
  if (isPopupItem(data)) {
    setPopupItem(data);
  } else {
    setPopupItem(undefined);
    focus();
    exec(editor);
    onClick?.(id);
  }
};
```

src/extensions/behavior/CommandMenu/component.tsx
```diff
@@ -76,17 +79,29 @@ export const CommandMenuComponent: React.FC<CommandMenuComponentProps> = ({
 function renderItem({id, title, icon, hotkey, hint}: CommandMenuItem): React.ReactNode {
     const titleText = isFunction(title) ? title() : title;
     const hintText = isFunction(hint) ? hint() : hint;
+    const preview = previews[id as keyof typeof ActionName];
```
makhnatkin, line 82: Now the previews are hardcoded in the code, we have to take them from the props, and we should take them from the view item. Add [another attribute](https://github.com/gravity-ui/markdown-editor/blob/main/src/modules/toolbars/types.ts#L22) to the view item and send the preview there, instead of using a single config as it is now.

src/bundle/config/previews/HeadingPreview.tsx
```diff
@@ -0,0 +1,22 @@
...
+import {i18n} from '../../../i18n/action-previews';
+
+import {ActionPreview} from './ActionPreview';
+
+type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6;
+
+const DynamicHeading = ({level, children}: {level: HeadingLevel; children: React.ReactNode}) => {
+    const HeadingTag = `h${level}`; // e.g. 'h1', 'h2', etc.
+
+    return React.createElement(HeadingTag, {}, children);
```
makhnatkin, line 12: Lets avoid createElement

use

```
const CustomTag = `h${this.props.level}`;

<CustomTag>Hello</CustomTag>
```

if you will have issues with TS use:
```
// var name must start with a capital letter
const CustomTag = `h${this.props.level}` as keyof JSX.IntrinsicElements;
```

src/toolbar/ToolbarListButton.scss
```diff
@@ -29,6 +29,17 @@
...
+        padding: var(--g-spacing-3);
+
+        border-radius: var(--g-border-radius-s);
+        background-color: var(--g-color-base-float-heavy);
+
+        & .g-popup__content[class] {
+            padding: 0 0 0 var(--g-spacing-2);
+        }
+    }
+
```
makhnatkin, line 42: why this style is duplicate?

src/toolbar/ToolbarListButton.scss
```diff
@@ -29,6 +29,17 @@
         }
     }
 
+    &__preview-content {
+        padding: var(--g-spacing-3);
+
+        border-radius: var(--g-border-radius-s);
+        background-color: var(--g-color-base-float-heavy);
+
+        & .g-popup__content[class] {
```
makhnatkin, line 38: it looks a little bit strange

src/i18n/action-previews/ru.json
```diff
@@ -0,0 +1,5 @@
+{
+  "text": "Это текст без заголовка. И заголовок, и \nтекст под ним можно выделять \nполужирным, курсивом, цветом, \nзачёркивать и подчёркивать. А ещё — \nдобавлять списки, таблицы, ссылки, \nформулы, якоря и блоки кода",
+  "text-with-head": "Это текст с заголовком. И заголовок, и \nтекст под ним можно выделять \nполужирным, курсивом, цветом, \nзачёркивать и подчёркивать. А ещё — \nдобавлять списки, таблицы, ссылки, \nформулы, якоря и блоки кода.",
+  "heading": "О тексте"
```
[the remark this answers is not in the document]
vladKrk: These are already proofread texts, and would not be fully visible, like this
image
So it is more like placeholder

src/toolbar/ToolbarListButton.scss
vladKrk: Create PreviewTooltip component and remove duplicate styles

src/toolbar/ToolbarListButton.scss
vladKrk: Tooltip has no options to remove inner paddings
It's for more specific class, can replace with more understanable :first-child

src/toolbar/ToolbarListButton.tsx
```diff
@@ -4,6 +4,7 @@ import {HelpPopover} from '@gravity-ui/components';
 import {ChevronDown} from '@gravity-ui/icons';
 import {ActionTooltip, Button, Hotkey, Icon, Menu, Popover, Popup} from '@gravity-ui/uikit';
 
+import {PreviewTooltip} from '../bundle/config/previews/PreviewTooltip';
```
makhnatkin, line 7: move PreviewTooltip from `src/bundle` close to  `src/toolbar/ToolbarListButton.tsx`

vladKrk: Why this particular place? PreviewTooltip is also used in src/extensions/behavior/CommandMenu/component.tsx

d3m1d0v (approved):

PR #586, opened 2025-02-05
feat(build)!: use new jsx transform

d3m1d0v:
For more info see [Introducing the New JSX Transform](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)


src/extensions/additional/GPT/IconRefuge/IconRefuge.tsx
```diff
@@ -20,7 +17,7 @@ import './IconRefuge.scss';
  * refugeWidth and refugeHeight
  */
 
-export const IconRefuge: FC<IconRefugeProps> = ({
+export const IconRefuge: React.FC<IconRefugeProps> = ({
```
makhnatkin, line 20: nit:

`import type {FC} from 'react';` It seems to be more in line with the style of all the other places. No?

src/plugins/TooltipButton/TooltipButton.tsx
```diff
@@ -1,4 +1,4 @@
-import React, {ReactNode, useLayoutEffect, useRef, useState} from 'react';
+import {type ReactNode, useLayoutEffect, useRef, useState} from 'react';
```
makhnatkin, line 1: maybe like this?

```ts
import type {ReactNode} from 'react';
import {useLayoutEffect, useRef, useState} from 'react';
```

src/view/hocs/withLatex/index.tsx
```diff
@@ -1,4 +1,4 @@
-import React, {ComponentType, RefAttributes, forwardRef, useEffect} from 'react';
+import {type ComponentType, type RefAttributes, forwardRef, useEffect} from 'react';
```
makhnatkin, line 1: maybe like this?
```
import type {ComponentType, RefAttributes} from 'react';
import {forwardRef, useEffect} from 'react';
```

makhnatkin (approved):

src/plugins/TooltipButton/TooltipButton.tsx
d3m1d0v: Fixed after #587

PR #639, opened 2025-02-18
feat(ImgSize): added ability to render custom form in image widget

d3m1d0v:
See `Examples / Custom Image Widget` page in storybook

For #624

### DIFF
--- a/demo/stories/examples/img-custom-form/ImageCustomForm.stories.tsx
+++ b/demo/stories/examples/img-custom-form/ImageCustomForm.stories.tsx
@@ -1,0 +1,13 @@
+import type {StoryObj} from '@storybook/react';
+
+import {ImageCustomFormDemo as component} from './ImageCustomForm';
+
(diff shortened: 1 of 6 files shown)


adambirds:
Hi @d3m1d0v, won't let me approve like you asked, but this is great. I just tried out the example in the playground, and its exactly what I was after. I will probably do what you said and merge it with the current one so that adding an image by link is still possible too. Thanks for doing this so quickly.

adambirds:
Will that example and the source code stay there in the playground now so I can look back on it when I get to this in a week or so? @d3m1d0v

d3m1d0v:
> Will that example and the source code stay there in the playground now so I can look back on it when I get to this in a week or so? @d3m1d0v

Yes, entire code from current PR will be merged into `main` branch and current example, together with source code, will be available

makhnatkin (approved):

PR #646, opened 2025-02-20
feat: new logger implementation

d3m1d0v:
Problems of old logger:
- global logger instance
  - there is no inheritance/nesting of logs
  - it is not possible to split logs between different instances of the editor
- only one log consumer
  - adding new consumer through `logger.setLogger` always override previous consumer

New logger:
- one per editor instance
- can be passed from outside
- support nesting
- event-based system
  - can add multiple subscribers to each log event

An alias named `globalLogger` has been added for old logger, both (`logger` and `globalLogger`) have been deprecated.

New logger is named `Logger2`. New instance can be created using `new Logger2()`. Nested logger can be created by calling `logger.nested({})`. Example: `new Logger2().nested({env: 'test'})`. Parameters passed to `.nested()` call will be added to all logs emitted by nested logger.

### DIFF
--- a/src/utils/facet.ts
+++ b/src/utils/facet.ts
@@ -1,0 +1,46 @@
+// Partial realization of Facet from CodeMirror, but for ProseMirror
+
+import {type EditorState, Plugin, PluginKey} from 'prosemirror-state';
+
+import {uniqueId} from '../lodash';
+
+/** @internal @unstable */
+export type FacetConfig<Input, Output> = {
+    combine: (value: readonly Input[]) => Output;
+    static: true;
+    name?: string;
+};
+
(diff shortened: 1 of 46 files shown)


src/utils/facet.ts
```diff
@@ -0,0 +1,46 @@
+// Partial realization of Facet from CodeMirror, but for ProseMirror
+
+import {type EditorState, Plugin, PluginKey} from 'prosemirror-state';
+
+import {uniqueId} from '../lodash';
+
+/** @internal */
```
makhnatkin, line 7: lets also add `@unstable`

makhnatkin (approved):

PR #648, opened 2025-02-20
fix(link): fixed link cursor focus

ReFFaT:
I solved a problem where if we used a link to the selected text, the focus would be on the penultimate character. The problem was that the link wraps the selected text in [text] and adds one character before. It turns out that the focus that was already there remains in place, but an invisible additional character appears in the text. I solved the problem by adding an empty text selection. However, it was not possible to attach it to the existing dispatch — the link was copied to the plain text. When creating a new dispatch, the problem was solved. If you can tell me how to get around this, I would fix the re-dispatch.


src/extensions/markdown/Link/plugins/LinkTooltipPlugin/index.tsx
```diff
@@ -239,6 +239,8 @@ class SelectionTooltip implements PluginView {
                 view.state.tr.addMark(from, to, linkType(view.state.schema).create({href: url})),
             );
 
+            view.dispatch(view.state.tr.setSelection(TextSelection.create(view.state.doc, to)));
```
makhnatkin, line 242: 1. use this code
```
        if (normalizeResult) {
            const {url} = normalizeResult;
            const {from, to} = textNode;

            const tr = view.state.tr;
            tr.setSelection(TextSelection.create(tr.doc, tr.mapping.map(to)));
            tr.addMark(from, to, linkType(view.state.schema).create({href: url}));
            view.dispatch(tr);
        }
```
2. you no need `setTimeout` more
3. this logic in update, removePlaceholderLink

src/extensions/markdown/Link/plugins/LinkTooltipPlugin/index.tsx
```diff
@@ -111,14 +111,12 @@ class SelectionTooltip implements PluginView {
...
             return;
 
-        const prevTextNode = this.textNode;
         this.textNode = getTextNode(view.state);
 
         const prevRef = this.textNodeRef;
         this.updateTextNodeRef();
 
         if (!this.textNode || !this.textNodeRef) {
-            this.removePlaceholderLink(prevTextNode);
```
makhnatkin, line 121: 1. Select the word
2. Click the link button
3. A popup will appear
4. Click somewhere on the document

Previously
the link was deleted because the text was not entered in the popup

Now
the link is not deleted

You should not change this behavior.

PR #650, opened 2025-02-21
feat: add more event logs

d3m1d0v:


src/extensions/behavior/FilePaste/index.ts
```diff
@@ -36,6 +42,14 @@ export const FilePaste: ExtensionAuto = (builder) => {
                     if (dropPos === -1) return false;
 
                     const posToInsert = dropPoint(view.state.doc, dropPos, slice);
+
+                    getLoggerFromState(view.state).event({
+                        event: 'drop-files',
+                        plugin: 'file-paste',
+                        prop: 'handleDrop',
```
makhnatkin, line 49: nit
its looks like stack trace info, but you need add it by hands

makhnatkin (approved):

PR #658, opened 2025-02-27
feat(Clipboard): added trim empty list items before copying

makhnatkin:


src/extensions/behavior/Clipboard/utils.ts
```diff
@@ -56,3 +58,57 @@ export function extractTextContentFromHtml(html: string) {
...
+                    let hasContent = false;
+
+                    node.content.forEach((child) => {
+                        if (
+                            (child.isText && child.text?.trim().length) ||
+                            (!child.isText && child.content.size > 0)
+                        ) {
+                            hasContent = true;
+                        }
+                    });
```
d3m1d0v, line 83: u can use `isEmptyString` from `BaseSchemaSpecs` instead

src/extensions/behavior/Clipboard/utils.ts
```diff
@@ -56,3 +61,48 @@ export function extractTextContentFromHtml(html: string) {
...
+    return [firstNotEmptyNodePos, lastNotEmptyNodePos, firstNodePos, lastNodePos];
+}
+
+export function trimContent(fragment: Fragment): Fragment {
+    const [notEmptyStart, notEmptyEnd, start, end] = findNotEmptyContentPosses(fragment);
+
+    if (notEmptyStart === start && notEmptyEnd === end) {
+        return fragment;
+    } else if (notEmptyStart === -1 && notEmptyEnd === -1) {
+        return fragment.cut(0, 1);
```
shevnv, line 104: I don't understand logic inside if statement `notEmptyStart === -1 && notEmptyEnd === -1`. Could you add some details here?

shevnv (approved):

PR #677, opened 2025-03-11
feat(wysiwyg): use shared state to manage editing state in mermaid and html-block nodes

d3m1d0v:
The `Mermaid` and `YfmHtmlBlock` extensions use `newCreated` attributes to automatically change state to editing after the node is created. This works fine for single editing.

But for collaborative editing, this causes all these nodes to be in editing state after users join the session. And this is very annoying.

I fixed this behavior by using shared state to manage editing state for these nodes.

**How it works:**

You need to generate an entity id before creating a new node. And add metadata to the transaction: pass the generated entity id and the initial state for node (`editing:true`). This metadata will be handled by the `SharedState` plugin and store the initial state for node inside plugin's own state.

After that in nodeview we can get the state for the current node from shared state using entity id from node attributes and go to the edit state of node if needed.

This shared state is only available to the current client in collaborative session and is not available to others. So the node will go to edit state only for the current user.


src/utils/entity-id.ts
```diff
@@ -0,0 +1,5 @@
+import {uniqueId} from 'src/lodash';
```
makhnatkin, line 1: nit
`uniqueId` is just a sequential number, maybe `uuid` is better?

d3m1d0v: yeah, changed to uuid v4

src/react-utils/useSharedEditingState.ts
```diff
@@ -0,0 +1,53 @@
...
+                        view.state.tr,
+                        // @ts-expect-error TS 2345 // TODO: fixme
+                        {editing: !valueRef.current},
+                    ),
+                );
+            },
+        };
+    }, [key, view, valueRef]);
+
+    return [value, set, unset, toggle] as const;
```
makhnatkin, line 28: I propose to simplify the structure
```ts
const createEditingDispatch = (view, key, editing) => view.dispatch(
      key.appendTransaction.merge(
          view.state.tr,
          // @ts-expect-error TS 2345 // TODO: fixme
          {editing},
      ),
  );

// ...

const {set, unset, toggle} = useMemo(() => {
  set: createEditingDispatch(view, key, true)
  // ...
});

```

src/utils/descedants.ts
```diff
@@ -0,0 +1,25 @@
+import type {Node} from '#pm/model';
+
+export type FindDescedantItem = {
+    node: Node;
+    pos: number;
+    index: number;
+    parent: Node | null;
+};
+
+export function findFirstDescedantNode(
```
makhnatkin, line 10: I suggest adding TSDoc

src/extensions/behavior/SharedState/SharedState.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import {pluginSpec} from './plugin';
+
+export {SharedStateKey, type SharedStateKeySpec} from './utils';
+
+/**
+ * This extension enables a plugin that stores internal state and handles its changes via meta-information in transactions.
+ * You can get, change and subscribe to state changes using a SharedStateKey instance.
+ */
+export const SharedState: Extension = (builder) => {
```
makhnatkin, line 12: I suggest adding TSDoc and brief information to the documentation about what can be used for any editor extensions

Perhaps it is worth adding a mention of the similarity of the idea with codemirror facets

makhnatkin (approved):

PR #721, opened 2025-04-16
fix(markup): fix pasting image urls to link lines

kseniya57:


src/markup/codemirror/create.ts
```diff
@@ -231,7 +233,7 @@ export function createCodemirror(params: CreateCodemirrorParams) {
                     }
                 }
 
-                if (parseInsertedUrlAsImage) {
+                if (parseInsertedUrlAsImage && !linkLineRegex.test(currentLine)) {
```
makhnatkin, line 236: Am I correct in understanding that this condition checks whether the currentLine is not a plain Markdown link ([text](url)), and if so—meaning the line contains just a pasted URL rather than a formatted link—then we should try to treat it as an image, provided that the parseInsertedUrlAsImage flag is enabled?

d3m1d0v (approved):

PR #729, opened 2025-04-21
feat(core): added preserving and managing original markdown formatting for hidden comments

saxumcordis:


src/bundle/config/dynamicModifiers.ts
```diff
@@ -24,13 +24,25 @@ export function createDynamicModifiers(markupManager: MarkupManager): DynamicMod
...
+        {
+            type: 'parserToken',
+            tokenName: 'hidden_comment',
+            process: (token, _, rawMarkup) => {
+                const tokenId = v5(rawMarkup, markupManager.getNamespace());
+
+                token.attrSet(YFM_TOKEN_ATTR, tokenId);
+                markupManager.setMarkup(tokenId, dedent(rawMarkup));
+
+                return token;
```
makhnatkin, line 43: Let's extract the repeated logic below into a utility function

```ts
const tokenId = v5(rawMarkup, markupManager.getNamespace());

token.attrSet(YFM_TOKEN_ATTR, tokenId);
markupManager.setMarkup(tokenId, dedent(rawMarkup));

return token;
```

makhnatkin: Consider refactoring the "enabled" flag into an array of allowed token types.

This would give us more flexibility and a safer default behavior. Instead of a simple boolean,
we could support an array like `allowed: ['yfm_table', 'yfm_cut']`.

If the array is **not** provided, we treat all as disallowed by default (safer assumption).
Perhaps, in the logic, we'd need a mapping like `yfm_table` → `yfm_table_open`, `yfm_table_close`, etc.

This matches the pattern used in the `disableTableModification` experiment —
see the storybook example: https://preview.gravity-ui.com/md-editor/?path=/story/experiments--story

makhnatkin:
Consider refactoring the "enabled" flag into an array of allowed token types.

This would give us more flexibility and a safer default behavior. Instead of a simple boolean,
we could support an array like `allowed: ['yfm_table', 'yfm_cut']`.

If the array is **not** provided, we treat all as disallowed by default (safer assumption).
Perhaps, in the logic, we'd need a mapping like `yfm_table` → `yfm_table_open`, `yfm_table_close`, etc.

This matches the pattern used in the `disableTableModification` experiment —
see the storybook example: https://preview.gravity-ui.com/md-editor/?path=/story/experiments--story

src/bundle/types.ts
```diff
@@ -119,9 +119,11 @@ export type MarkdownEditorExperimentalOptions = {
      *
      * **Note:** Applies to tracked blocks: `yfm_table`
      *
-     * @default false
+     * **Note:** Preserve formatting by prefix: ['yfm_table'] works for `yfm_table_open` too
+     *
+     * @default []
      */
-    preserveMarkupFormatting?: boolean;
+    preserveMarkupFormattingFeatures?: string[];
```
d3m1d0v, line 126: breaking changes...

src/presets/full-specs.ts
```diff
@@ -19,6 +20,8 @@ export const FullSpecsPreset: ExtensionAuto<FullSpecsPresetOptions> = (builder,
 
     builder.use(MarkSpecs).use(ColorSpecs, opts.color);
 
+    builder.use(HiddenCommentBlock);
```
d3m1d0v, line 23: why should this be part of full-preset?

saxumcordis:
Fixed on service side by using plugin

PR #732, opened 2025-04-22
feat(SelectionContext): allow to change popup placement behavior

d3m1d0v:
See [story](https://preview.gravity-ui.com/md-editor/732/?path=/story/settings-wysiwyg-selection-context--story)


d3m1d0v:
> Should we add visual tests for flip behavior?

I suggest adding tests for the entire behavior of selection context, but in a separate issue/pr

PR #736, opened 2025-04-23
feat(markup): allow to disable search panel

DaniilDem:


src/bundle/types.ts
```diff
@@ -163,6 +163,10 @@ export type MarkdownEditorMarkupConfig = {
      * The function, used to determine if the pasted text is the image url and should be inserted as an image
      */
     parseInsertedUrlAsImage?: ParseInsertedUrlAsImage;
+    /**
+     * Disable the search panel in the editor.
+     */
+    disableSearchPanel?: boolean;
```
d3m1d0v, line 169: lets rename to `searchPanel?: boolean`

and add default value to jsdoc

DaniilDem: do you want `searchPanel` to be `true` by default?

d3m1d0v: > do you want `searchPanel` to be `true` by default?

yes

d3m1d0v (approved):

d3m1d0v:
👍

PR #746, opened 2025-04-30
feat(core, color): added support for custom escape characters

makhnatkin:
The `esc()` method now checks `state.escapeCharacters`, if defined, and includes its characters in the escape regex.  
This allows context-specific escaping — for example, in the `color` mark, to escape parentheses in user input without interfering with the markup syntax.

### DIFF
--- a/src/core/ExtensionsManager.ts
+++ b/src/core/ExtensionsManager.ts
@@ -129,7 +129,7 @@ export class ExtensionsManager {
         this.#spec.marks().forEach(this.processMark);
     }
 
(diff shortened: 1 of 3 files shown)


makhnatkin:
https://github.com/nkjmsss/markdown-it-color/pull/17

makhnatkin:
Found one issue — in markup mode, using the toolbar button doesn’t add escaping. Since we work with plain text there, I’m not sure how easy it is to support


https://github.com/user-attachments/assets/b8102fd1-93cd-473e-a7aa-17f3935acf42

PR #751, opened 2025-05-13
fix(styles): added spacing rules to prevent jumps when editing lists

makhnatkin:
See also https://github.com/diplodoc-platform/transform/pull/708


d3m1d0v (approved):

d3m1d0v (approved):

src/styles/list.scss
```diff
@@ -0,0 +1,17 @@
+/*
+ TDOD: delete it after https://github.com/diplodoc-platform/transform/pull/708
```
makhnatkin, line 2: fix: TODO

PR #781, opened 2025-06-08
fix(GPT extension): fixed opened gpt in safari brouse

ReFFaT:
Here is the corrected text:

I managed to fix the extension, I realized what the problem was.
There are two points there:

Safari browser losing focus when opening a popup, which caused the close function on the popup to trigger, causing the popup to close.

The problem was still in the plugin in WYSIWYG mode. In Safari browser for some reason after triggering the update function, where there is this.rerender(), gptWidgetPlugin state -> init was also triggered, but this is only in Safari browser. And since the meta data was empty, the selection was set to 0 0 by default, so after the update was triggered again and compared the current selection with the previous one and since they were different, it just closed the popup. I added logic to the plugin where it returns the current state when the meta is empty, and removed the code in update that was responsible for closing the popup. Why close it in update when the plugin component does it. But maybe I could have broken something, I poked around the main cases, everything seems fine, can you look at the cases too?

Translated with DeepL.com (free version)

### DIFF
--- a/src/extensions/additional/GPT/MarkupGpt/popup.tsx
+++ b/src/extensions/additional/GPT/MarkupGpt/popup.tsx
@@ -21,8 +21,10 @@ export function renderPopup<
             className={cnGptPopup()}
             open
             strategy="absolute"
-            onOpenChange={(open) => {
-                if (!open) props.onClose();
+            onOpenChange={(_open, _event, reason) => {
+                if (reason === 'outside-press' || reason === 'escape-key') {
+                    props.onClose();
+                }
             }}
             anchorElement={anchor}
             placement={gptPopupPlacement}
--- a/src/extensions/additional/GPT/gptExtension/view.tsx
+++ b/src/extensions/additional/GPT/gptExtension/view.tsx
@@ -3,7 +3,7 @@ import {useCallback} from 'react';
 import {Popup} from '@gravity-ui/uikit';
 import type {PopupProps} from '@gravity-ui/uikit';
 import {Slice} from 'prosemirror-model';
-import type {EditorState, PluginView} from 'prosemirror-state';
+import type {PluginView} from 'prosemirror-state';
 import {TextSelection} from 'prosemirror-state';
 import type {EditorView} from 'prosemirror-view';
 import {useMount} from 'react-use';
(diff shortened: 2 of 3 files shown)


src/extensions/additional/GPT/gptExtension/view.tsx
```diff
@@ -152,6 +140,8 @@ export class GptWidgetDecoView<
         const tr = this._view.state.tr;
         const meta: GptWidgetMeta = {action: 'hide'};
 
+        console.log('submit meta', meta);
```
makhnatkin: delete it plz

src/extensions/additional/GPT/gptExtension/view.tsx
```diff
@@ -177,6 +167,8 @@ export class GptWidgetDecoView<
         const tr = this._view.state.tr;
         const meta: GptWidgetMeta = {action: 'hide'};
 
+        console.log('close meta', meta);
```
makhnatkin: delete it plz

src/extensions/additional/GPT/plugin.ts
```diff
@@ -35,6 +35,8 @@ export const gptWidgetPlugin = <
                 const meta = tr.getMeta(key) as GptWidgetMeta | undefined;
                 const paramsGpt = params;
 
+                if (!meta) return decos;
```
makhnatkin: Could you explain why this code was added?
	1.	It seems unnecessary.
	2.	It might break the logic, as it return decos without mapping.

src/extensions/additional/GPT/MarkupGpt/popup.tsx
```diff
@@ -21,8 +21,10 @@ export function renderPopup<
             className={cnGptPopup()}
             open
             strategy="absolute"
-            onOpenChange={(open) => {
-                if (!open) props.onClose();
+            onOpenChange={(_open, _event, reason) => {
+                if (reason === 'outside-press' || reason === 'escape-key') {
+                    props.onClose();
+                }
```
ReFFaT: I added a handler for certain actions, since the Safari browser by default fires this event along with the renderer, I don't know why it thinks it's losing focus and triggers it

src/extensions/additional/GPT/gptExtension/view.tsx
```diff
@@ -281,8 +267,10 @@ function Widget<
                 open
                 anchorElement={anchorElement}
                 placement={gptPopupPlacement}
-                onOpenChange={(open) => {
-                    if (!open) onClose();
+                onOpenChange={(_open, _event, reason) => {
+                    if (reason === 'outside-press' || reason === 'escape-key') {
+                        onClose();
+                    }
```
ReFFaT: same as above

src/extensions/additional/GPT/plugin.ts
```diff
@@ -35,9 +35,9 @@ export const gptWidgetPlugin = <
                 const meta = tr.getMeta(key) as GptWidgetMeta | undefined;
                 const paramsGpt = params;
 
-                paramsGpt.disablePromptPresets = false;
-
                 if (meta?.action === 'show') {
+                    paramsGpt.disablePromptPresets = false;
+
```
ReFFaT, line 40: I moved it inside because when I run the apply function again, the gpt mode changes from single selection to intermediate selection, so this problem won't occur.

d3m1d0v (approved):

makhnatkin (approved):

PR #783, opened 2025-06-11
feat: added data attribute for searching toolbar button

kkirik:


src/toolbar/ToolbarButton.tsx
```diff
@@ -69,6 +69,7 @@ export const ToolbarButtonView = forwardRef<HTMLButtonElement, ToolbarButtonView
                                 onClick={onClick}
                                 className={b(null, [className])}
                                 aria-label={titleText}
+                                data-toolbar-button-name={titleText}
```
makhnatkin: 1. There's probably no need to duplicate `aria-label` in `data-toolbar-button-name`.
2. `titleText` is language-dependent, so it's better to store a unique identifier in the data attribute — for example, the `id` from the toolbar configuration.
3. Suggest using a more generic and descriptive name for the data attribute, such as `data-toolbar-item`.

kkirik: Didn't see that this component has an id prop

src/toolbar/ToolbarButton.tsx
```diff
@@ -72,6 +85,7 @@ export const ToolbarButtonView = forwardRef<HTMLButtonElement, ToolbarButtonView
                                 onClick={onClick}
                                 className={b(null, [className])}
                                 aria-label={titleText}
+                                data-toolbar-item={id}
```
makhnatkin, line 88: @d3m1d0v 
We want to provide a way to reference the toolbar button via an attribute. 
What if approach might look like this:

```
    return (
        <ToolbarButtonView
            {...props}
            id={`toolbar-item-${id}`}
            qa={qa ?? `toolbar-item-${id}`}
            active={active}
            enabled={enabled}
            onClick={handleClick}
        />
    );
```

PR #784, opened 2025-06-11
test: added Colorify visual tests

makhnatkin:
Targeted tests have been implemented for the `Colorify` feature, supported by a dedicated utility class designed to streamline the testing process. Toolbar selectors were refined to improve both precision and maintainability.


tests/playwright/core/editor.ts
```diff
@@ -185,6 +185,24 @@ export class MarkdownEditorPage {
         await this.expect(button).not.toHaveClass(/selected/);
     }
 
+    /**
+     * Asserts that the toolbar color button has the "default" qa attribute.
+     */
+    async assertToolbarColorButtonDefault(inPopup = false) {
+        const root = inPopup ? this.page.locator('.g-popup.g-popup_open') : this.locators.editor;
+        const button = root.getByLabel('Text color');
+        await this.expect(button).toHaveAttribute('data-qa', /colors-default/);
```
makhnatkin: add other attr for state plz

tests/playwright/core/editor.ts
```diff
@@ -185,6 +185,24 @@ export class MarkdownEditorPage {
...
+
+    /**
+     * Asserts that the toolbar color button does not have the "default" qa attribute.
+     */
+    async assertToolbarColorButtonNotDefault(inPopup = false) {
+        const root = inPopup ? this.page.locator('.g-popup.g-popup_open') : this.locators.editor;
+        const button = root.getByLabel('Text color');
+        await this.expect(button).not.toHaveAttribute('data-qa', /colors-default/);
+    }
+
```
makhnatkin: rescreen https://github.com/gravity-ui/markdown-editor/blob/d93964ea0d520fe0ae4f709365e5592499e7e3c4/tests/visual-tests/playground/Toolbar.visual.test.tsx#L30

tests/visual-tests/playground/Colorify.visual.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+            await editor.focus();
+            await editor.press('ArrowDown');
+            await editor.press('Enter');
+
+            await editor.clickToolbarButton('Text color');
+            await wait.timeout();
+
+            await editor.clickToolbarButton('Yellow', true);
+            await wait.timeout();
```
makhnatkin, line 50: add selectInList

makhnatkin: The changes are scheduled to be introduced in the next pull request focused on the Heading feature.

src/bundle/toolbar/custom/ToolbarColors.tsx
```diff
@@ -83,6 +83,9 @@ export const ToolbarColors: React.FC<ToolbarColorsProps> = (props) => {
             _selectionType="light"
             iconClassName={b('menu-icon', {color: currentColor})}
             chevronIconClassName={b('chevron-icon', {color: currentColor})}
+            qa={`g-md-colorify g-md-colorify-${
+                currentColor && currentColor !== 'default' ? currentColor : 'default'
+            }`}
```
d3m1d0v: Attribute `qa` does not a classname, so let's not pass multiple values ​​at once. Current color can be stored in a custom `data-*` attribute

src/toolbar/FlexToolbar.tsx
```diff
@@ -64,7 +64,6 @@ export function FlexToolbar<E>(props: FlexToolbarProps<E>) {
                 {dots?.length && (
                     <ToolbarListButton
                         qa="g-md-toolbar-more-action"
-                        qaMenu="g-md-toolbar-more-menu"
```
d3m1d0v, line 67: ???

src/toolbar/ToolbarListButton.tsx
```diff
@@ -87,7 +84,7 @@ export function ToolbarListButton<E>({
                 {buttonContent}
             </ToolbarButtonView>
             <Popup anchorElement={anchorElement} open={popupOpen} onOpenChange={hide}>
-                <Menu size="l" className={b('menu')} qa={qaMenu}>
+                <Menu size="l" className={b('menu')} qa="g-md-toolbar-additional">
```
d3m1d0v: `qa` attribute for this Menu should be set from props, since it is a generic component and is used for other lists as well.

d3m1d0v (changes requested):

src/toolbar/ToolbarListButton.tsx
makhnatkin: 1. Multiple `qa` attributes usually signal that the component should be broken down into smaller parts.
2. I’d prefer to keep `qa` usage minimal. Still, if we decide to restore it, we should also introduce a `qaActionDisabledPopover`.
3. Although it’s designed as a generic component, it’s currently used in just one place.

d3m1d0v: > 3\. Although it’s designed as a generic component, it’s currently used in just one place.

Currently Headings, Lists, Code and Math toolbar items are rendered using this component

src/bundle/toolbar/custom/ToolbarColors.tsx
makhnatkin: Yes, I’ve considered this and agree with the argument. At the same time, it’s important to note that I haven’t come up with a good abstract name for this data attribute — `data-whatToPutHere`?
That’s because `ToolbarButtonWithPopupMenu` is a generic component and isn’t used only for colors.
So we probably need to add something like a `data-additional-info` attribute.

src/toolbar/FlexToolbar.tsx
makhnatkin: Reverted because `Headings`, `Lists`, `Code`, and `Math` toolbar items currently rely on `ToolbarListButton` as well.

src/bundle/toolbar/ToolbarButtonWithPopupMenu.tsx
```diff
@@ -39,6 +39,7 @@ export type ToolbarButtonWithPopupMenuProps = Omit<
             chevronIconClassName?: string;
             title: string | (() => string);
             menuItems: MenuItem[];
+            additionalInfo?: string;
```
d3m1d0v: unused prop

d3m1d0v (approved):

PR #800, opened 2025-07-02
feat(YfmTable): support for ignoring splitters in table content

d3m1d0v:
see diplodoc-platform/transform#724


src/extensions/yfm/YfmTable/YfmTableSpecs/index.ts
```diff
@@ -11,13 +12,39 @@ import {serializerTokens} from './serializer';
 export {YfmTableNode} from './const';
 export {yfmTableType, yfmTableBodyType, yfmTableRowType, yfmTableCellType} from './utils';
 
-export type YfmTableSpecsOptions = YfmTableSchemaOptions & {};
+export type YfmTableSpecsOptions = YfmTableSchemaOptions &
+    Pick<
+        YfmTablePluginOptions,
+        | 'table_ignoreSplittersInBlockCode'
```
makhnatkin, line 18: I suggest not mixing snake_case and camelCase, and using camelCase like this: `tableIgnoreSplittersInBlockMath`

PR #822, opened 2025-07-29
feat: export Math extension

Coudnet:
To import Math extension like GPT extension directly from `@gravity-ui/markdown-editor`

`import { Math } from '@gravity-ui/markdown-editor';`


makhnatkin:
We cannot export Math extension because right now [latex](https://github.com/gravity-ui/markdown-editor/blob/main/package.json#L302) a peer dependency

Issue #825, opened 2025-08-06
Do not replace ... with a Unicode symbol in WYSIWYG mode


kkirik:
If you type `text...` in WYSIWYG mode, the three dots are replaced with the Unicode symbol …

PR #829, opened 2025-08-11
fix: ellipsis rule in inline code

obenjiro:
## Summary
- avoid converting `...` to ellipsis within inline code
- add tests ensuring ellipsis still works in regular text

Fixes https://github.com/gravity-ui/markdown-editor/issues/825


src/extensions/base/BaseInputRules/index.ts
d3m1d0v, line 1: This issue has been fixed in `prosemirror-inputrules` since [v1.5.0](https://github.com/ProseMirror/prosemirror-inputrules/blob/master/CHANGELOG.md#150-2025-03-18)
So you just need to update `prosemirror-inputrules`

obenjiro: updated to prosemirror-inputrules

d3m1d0v (approved):

PR #831, opened 2025-08-18
feat(YfmTable): rework table controls – add drag-n-drop, rows and columns insertion

d3m1d0v:
- Controls for adding rows and columns have been redesigned:
  - the vertical and horizontal plus buttons at the right and bottom edges of the table have been removed;
  - instead, floating plus buttons have been added near the top and left edges of the table for each row and column;
    - hover over dots to show and click plus button;
- Menu controls for rows and columns have also been redesigned:
  - on hover over a row or column, the menu control will show a compact view. Hover over control to see a full-size button and icon;
  - drag the control to rearrange rows and columns;
  - сlick the control to open a dropdown-menu for the row or column;
  - a new action has been added to the dropdown-menu: "Clear cells".

### DIFF
--- a/src/extensions/yfm/YfmTable/index.ts
+++ b/src/extensions/yfm/YfmTable/index.ts
@@ -26,6 +26,13 @@ export type YfmTableOptions = YfmTableSpecsOptions & {
      * @default true
      */
     controls?: boolean;
+    /**
+     * Enables drag-d-drop rows and columns in table.
+     * The `controls` property must also be `true`.
+     *
+     * @default true
+     */
+    dnd?: boolean;
 };
 
 export const YfmTable: ExtensionWithOptions<YfmTableOptions> = (builder, options) => {
@@ -42,7 +49,7 @@ export const YfmTable: ExtensionWithOptions<YfmTableOptions> = (builder, options
     builder.addPlugin(yfmTableTransformPastedPlugin);
 
     if (options.controls !== false) {
-        builder.addPlugin(yfmTableControlsPlugins);
+        builder.addPlugin(yfmTableControlsPlugins({dndEnabled: options.dnd !== false}));
     }
 };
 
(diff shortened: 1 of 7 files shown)


src/table-utils/table-desc.ts
```diff
@@ -19,9 +19,25 @@ type TableCellVirtualDesc = {
...
+    safeTopBoundary: boolean;
+    safeBottomBoundary: boolean;
+};
+export type TableColumnRange = {
+    startIdx: number;
+    endIdx: number;
+    columnsCount: number;
+    safeLeftBoundary: boolean;
+    safeRightBoundary: boolean;
+};
```
makhnatkin, line 46: I suggest to add more comments, like:

TableDesc creates a virtual table description containing info about real and virtual cells needed for correct table operations.
This description is cached.
At the same time, document positions may change, so a separate class (Bounded) is added to work with specific positions.
Methods are provided for handling merged cells.
Ranges are also added for more comprehensive cell operations.

demo/defaults/content.ts
```diff
@@ -1,279 +1,353 @@
 export const markup = `
-&nbsp;
+{% cut "First" %}
```
makhnatkin: plz dont forget to return old markup

tests/visual-tests/__snapshots__/YfmTable.visual.test.tsx-snapshots/YfmTable-should-add-column-via-click-on-floating-plus-button-wysiwyg-dark-chromium-linux.png
makhnatkin, line 1: I suggest adding a fix for the background, and also bringing this fix to the diplodoc/transform package.

src/extensions/yfm/YfmTable/index.ts
```diff
@@ -19,7 +19,14 @@ export {
...
 } from './YfmTableSpecs';
 
-export type YfmTableOptions = YfmTableSpecsOptions & {};
+export type YfmTableOptions = YfmTableSpecsOptions & {
+    /**
+     * Enables floating controls for table.
+     *
+     * @default true
+     */
+    controls?: boolean;
```
makhnatkin, line 28: Previously, there were controls in the mobile version as well, but now they are only in the full version.

I suggest we think about how the controls should work in the mobile version.

PR #840, opened 2025-09-03
fix: correct toggle logic for partial selection

makhnatkin:
## Summary

Fix toggle behaviour for partial selections so toolbar actions and markdown shortcuts behave the same. Mixed inline-mark selections now apply the mark to the whole selection, fully covered selections still toggle off, and color keeps the expected ProseMirror whitespace and active-state semantics.

## What changed

- Inline markdown mark actions use `removeWhenPresent: false`, so mixed selections apply the mark instead of removing it.
- Markdown shortcuts for **Bold / Italic / Strike / Underline / Code** now use the same guarded command path as toolbar actions.
- `Color` now handles cursor and range selections explicitly, while preserving:
  - `isActive` for partially colored non-empty selections
  - whitespace trimming at selection boundaries when applying or replacing colors
  - toggle-off for fully covered same-color selections
- `WToolbarColors` no longer decides whether a click should clear the color; that logic now lives in the action.
- Markup `colorify` can unwrap the same color and replace another color when the selection exactly matches an existing `{color}(...)` wrapper.

## Known limitation

- Markup-mode wrapper detection is still intentionally narrow; broader existing-wrapper detection remains tracked in `#1129`.

## Tests

Added unit coverage for:

- mixed-selection toggle behaviour in `utils/actions.test.ts`
- command/keymap parity for markdown inline mark shortcuts
- removal on blocked punctuation when the selection is already fully covered by the mark
- `Color` cursor/range behaviour, including trailing-whitespace handling and partially colored selection state
- markup `colorify` unwrap/replace behaviour for exact wrapper selections


makhnatkin:
@sourcery-ai review

PR #849, opened 2025-09-29
feat(YfmTable): add ghost when dragging rows and columns

d3m1d0v:


src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd.ts
```diff
@@ -124,7 +125,7 @@ abstract class YfmTableDnDAbstractHandler implements TableHandler, DnDControlHan
         if (!this._dragMouseDown || !isDragThresholdPassed(this._dragMouseDown, event)) return;
 
         if (this._editorView.dragging || this._dragging) return;
-        this._startDragging();
+        this._startDragging(event);
```
makhnatkin, line 128: I’m getting an unstable grab of the row or column that I want to drag.
In the video below I performed the following actions:
	1.	I grabbed a row and tried to drag it upward, but it didn’t move.
	2.	I went back to grab it again, but instead of grabbing, the ghost immediately appeared.


https://github.com/user-attachments/assets/0dbe36eb-403a-46ec-b290-ac3e2407ba81

makhnatkin: Maybe don’t wait for debounce on the first step. Right after `this._startDragging(event)`, call the first DnD update synchronously, then use debounced updates.

src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd-ghost.ts
```diff
@@ -0,0 +1,226 @@
...
+            shiftY: rect.top - this._y,
+        };
+    }
+
+    private _buildGhostContainer(view: EditorView): HTMLElement {
+        const container = view.dom.ownerDocument.createElement('div');
+        const yfmClasses = view.dom.classList
+            .entries()
+            .map(([, val]) => val)
+            .filter((val) => val.startsWith('yfm_'));
```
makhnatkin: Instead of using `entries`, you can use `values`, or even shorter:

```suggestion
const yfmClasses = Array.from(view.dom.classList).filter((c) => c.startsWith('yfm_')); 
```

src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd-ghost.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    private _updatePositions() {
+        this._ghostTable.style.top = this._y + this._tblShiftY + 'px';
+        this._ghostTable.style.left = this._x + this._tblShiftX + 'px';
+
+        if (this._ghostButton) {
+            this._ghostButton.style.top = this._y + this._btnShiftY + 'px';
+            this._ghostButton.style.left = this._x + this._btnShiftX + 'px';
+        }
+    }
```
makhnatkin, line 109: I think using `transform` for the ghost’s motion is preferable to updating `top`/`left`.
`top`/`left` can trigger [layout](https://gist.github.com/faressoft/36cdd64faae21ed22948b458e6bf04d5) (reflow) and paint work, while `transform` changes are typically compositor-only (no layout/paint), so they run smoother and avoid main-thread jank.

maybe like this:
```
private _updatePositions() {
  const tx = this._x + this._tblShiftX;
  const ty = this._y + this._tblShiftY;

  this._ghostTable.style.transform = `translate3d(${tx}px, ${ty}px, 0)`;

  if (this._ghostButton) {
    const bx = this._x + this._btnShiftX;
    const by = this._y + this._btnShiftY;
    this._ghostButton.style.transform = `translate3d(${bx}px, ${by}px, 0)`;
  }
}
```

Also optional CSS tweak for best results:
```
.yfm.g-md-yfm-table-dnd-ghost,
.g-button.g-md-yfm-table-dnd-ghost-button {
  position: fixed;           /* keeps the ghost glued to viewport */
  will-change: transform;    /* pre-promote to a composited layer */
  pointer-events: none;      /* avoid intercepting pointer events */
}
```

References
https://paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/
https://web.dev/articles/animations-guide

src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd-ghost.ts
```diff
@@ -0,0 +1,226 @@
...
+            this._ghostTable = domElement;
+            this._tblShiftX = shiftX;
+            this._tblShiftY = shiftY;
+            this._dndBackgroundElem.appendChild(this._ghostTable);
+        }
+
+        this._updatePositions();
+
+        this._rafId = requestAnimationFrame(() => {
+            document.body.append(this._dndBackgroundElem);
```
makhnatkin, line 69: Earlier in this file you consistently use `ownerDocument`
Should this be `view.dom.ownerDocument.body.append(...)` instead of the global `document`?

src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd.ts
```diff
@@ -246,13 +250,18 @@ class YfmTableRowDnDHandler extends YfmTableDnDAbstractHandler {
...
 
+        const onMove = (event: MouseEvent) => {
+            ghost.move(event);
+            onMoveDebounced(event);
+        };
+
         document.addEventListener('mousemove', onMove);
 
         document.addEventListener(
             'mouseup',
```
makhnatkin, line 261: If I understood correctly, the mouseup listener is never removed, so it will accumulate across multiple drags. Each new drag attaches a new mouseup handler, leading to leaks and duplicate calls. Consider either [using](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#once) `{once: true}` or managing cleanup via `AbortController`.

makhnatkin: Should we also account for pointer events (so the logic works across mouse, touch, and pen inputs)? And hypothetically, what about the case when the window loses focus (blur) — should we end the drag in that situation?

src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd-ghost.ts
```diff
@@ -0,0 +1,226 @@
...
+        for (let rowIdx = range.startIdx; rowIdx <= range.endIdx; rowIdx++) {
+            const tr = tbody.appendChild(document.createElement('tr'));
+
+            for (let colIdx = 0; colIdx < tableDesc.cols; colIdx++) {
+                const cellPos = tableDesc.getPosForCell(rowIdx, colIdx);
+                if (cellPos.type === 'real') {
+                    const origNode = view.domAtPos(cellPos.from + 1).node as HTMLElement;
+                    const cloned = tr.appendChild(origNode.cloneNode(true));
+
+                    const rect = origNode.getBoundingClientRect();
```
makhnatkin, line 141: If I understood correctly, the current logic is: take the real table cell, clone it into the ghost, and immediately set its width/height. The issue is that the clone keeps all attributes (including id), so the DOM may end up with duplicates, and also mixing read (getBoundingClientRect) with write (style.width/height) can trigger extra reflows.

Maybe it makes sense to split this into phases: first read sizes, then clean up the clone, then write sizes. In other words: read everything first, then write everything.

example:
```
const origNode = view.domAtPos(cellPos.from + 1).node as HTMLElement;
const cloned = tr.appendChild(origNode.cloneNode(true)) as HTMLElement;

// read
const rect = origNode.getBoundingClientRect();

// clean
cloned.removeAttribute('id');
cloned.querySelectorAll('[id]').forEach((el) => el.removeAttribute('id'));
cloned.setAttribute('aria-hidden', 'true');

// write
cloned.style.width = rect.width + 'px';
cloned.style.height = rect.height + 'px';
```

d3m1d0v: > mixing read (getBoundingClientRect) with write (style.width/height) can trigger extra reflows

This can't trigger reflows, because cloned cell is attached to document later

d3m1d0v: > cloned.setAttribute('aria-hidden', 'true');

I think it’s better to set this attribute to container instead of setting it for each cell?

d3m1d0v: > > cloned.setAttribute('aria-hidden', 'true');
> 
> I think it’s better to set this attribute to container instead of setting it for each cell?

Similarly, cleaning `id` attributes

src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd.ts
d3m1d0v: > the mouseup listener is never removed

`{once: true}` already uses (line 268)

d3m1d0v: > Should we also account for pointer events (so the logic works across mouse, touch, and pen inputs)?

No, because this functionality is disabled when `mobile: true`

src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd.ts
d3m1d0v: This issue is not related to the addition of the ghost, so I will fix it in a separate PR.

makhnatkin (approved):

src/extensions/yfm/YfmTable/plugins/YfmTableControls/dnd/dnd.ts
d3m1d0v: #853

PR #857, opened 2025-10-05
fix(imageSize): fixed the insertion of the svg image

ReFFaT:
Fixed SVG image insertion via URL paste and copy/paste.

Problem: Browser renders SVG with 0x0 dimensions when width/height attributes are missing.

Solution:
- Added default dimensions 300x200 for SVG without size attributes  
- Preload SVG to extract actual dimensions before insertion
- Added 2-second timeout fallback for slow loading images

Note: SVG images are loaded twice (dimension detection + rendering), but the second load 
uses browser cache, making the performance impact negligible.

Fixes issue where SVG images rendered with 0x0 size in editor

### DIFF
--- a/src/extensions/yfm/ImgSize/ImagePaste/index.ts
+++ b/src/extensions/yfm/ImgSize/ImagePaste/index.ts
@@ -12,7 +12,8 @@ import type {FileUploadHandler} from '../../../../utils';
 import {clipboardUtils} from '../../../behavior/Clipboard';
 import {DataTransferType} from '../../../behavior/Clipboard/utils';
 import {ImageAttr, ImgSizeAttr, imageType} from '../../../specs';
-import {type CreateImageNodeOptions, isImageNode} from '../utils';
+import {DEFAULT_SVG_HEIGHT, DEFAULT_SVG_WIDTH} from '../const';
+import {type CreateImageNodeOptions, checkSvg, isImageNode} from '../utils';
 
 import {ImagesUploadProcess} from './upload';
 
@@ -82,14 +83,82 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
 
                                 e.preventDefault();
 
-                                const imageNode = imageType(view.state.schema).create({
-                                    src: imageUrl,
-                                    alt: title,
-                                });
+                                const isSvg = checkSvg(imageUrl);
 
-                                const tr = view.state.tr.replaceSelectionWith(imageNode);
-                                view.dispatch(tr.scrollIntoView());
-                                logger.event({event: 'paste-url-as-image'});
+                                if (isSvg) {
+                                    const img = new Image();
+                                    let loaded = false;
+
+                                    const createImageNode = (
+                                        width: number,
+                                        height: number,
+                                        logType: string,
+                                    ) => {
+                                        const imageNode = imageType(view.state.schema).create({
+                                            src: imageUrl,
+                                            alt: title,
+                                            [ImgSizeAttr.Width]: width,
+                                            [ImgSizeAttr.Height]: height,
+                                        });
+
+                                        const tr = view.state.tr.replaceSelectionWith(imageNode);
+                                        view.dispatch(tr.scrollIntoView());
+                                        logger.event({
+                                            event: 'paste-url-as-image',
+                                            type: logType,
+                                        });
+                                    };
+
+                                    const timeoutId = setTimeout(() => {
+                                        if (!loaded) {
+                                            createImageNode(
+                                                DEFAULT_SVG_WIDTH,
+                                                DEFAULT_SVG_HEIGHT,
+                                                'svg-timeout',
+                                            );
+                                        }
+                                    }, 2000);
+
+                                    img.onload = () => {
+                                        loaded = true;
+
+                                        clearTimeout(timeoutId);
+
+                                        const width =
+                                            img.naturalWidth > 0
+                                                ? img.naturalWidth
+                                                : DEFAULT_SVG_WIDTH;
+                                        const height =
+                                            img.naturalHeight > 0
+                                                ? img.naturalHeight
+                                                : DEFAULT_SVG_HEIGHT;
+
+                                        createImageNode(width, height, 'svg-with-dimensions');
+                                    };
+
+                                    img.onerror = () => {
+                                        loaded = true;
+
+                                        clearTimeout(timeoutId);
+
+                                        createImageNode(
+                                            DEFAULT_SVG_WIDTH,
+                                            DEFAULT_SVG_HEIGHT,
+                                            'svg-error-fallback',
+                                        );
+                                    };
+
+                                    img.src = imageUrl;
+                                } else {
+                                    const imageNode = imageType(view.state.schema).create({
+                                        src: imageUrl,
+                                        alt: title,
+                                    });
+
+                                    const tr = view.state.tr.replaceSelectionWith(imageNode);
+                                    view.dispatch(tr.scrollIntoView());
+                                    logger.event({event: 'paste-url-as-image'});
+                                }
 
                                 return true;
                             }
--- a/src/extensions/yfm/ImgSize/utils.ts
+++ b/src/extensions/yfm/ImgSize/utils.ts
@@ -34,6 +34,18 @@ export const createImageNode =
                 logger.error({error: err});
             }
         }
+
+        const isSvg = checkSvg(result.url) || file.type === 'image/svg+xml';
+
+        if (isSvg) {
+            const sizes = await loadImage(file).then(getImageSizeNew);
+            return imgType.create({
+                ...attrs,
+                [ImgSizeAttr.Width]: sizes.width,
+                [ImgSizeAttr.Height]: sizes.height,
+            });
+        }
+
         return imgType.create(attrs);
     };
 
@@ -63,4 +75,8 @@ export function getImageSizeNew({width, height}: HTMLImageElement): {
         imgMaxHeight: IMG_MAX_HEIGHT,
     });
     return {width: String(size.width), height: String(size.height)};
+}
+
+export function checkSvg(imageUrl?: string) {
+    return imageUrl && (/\.svg($|\?|#)/i.test(imageUrl) || imageUrl.startsWith('data:image/svg'));
 }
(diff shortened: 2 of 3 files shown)


src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -82,14 +83,82 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
-                                const imageNode = imageType(view.state.schema).create({
-                                    src: imageUrl,
-                                    alt: title,
-                                });
+                                const isSvg = checkSvg(imageUrl);
 
-                                const tr = view.state.tr.replaceSelectionWith(imageNode);
-                                view.dispatch(tr.scrollIntoView());
-                                logger.event({event: 'paste-url-as-image'});
+                                if (isSvg) {
```
makhnatkin: The SVG code uses callback hell instead of promises, duplicates node creation logic 3 times, and delays image insertion until load/timeout. It doesn't reuse existing utilities [loadImage()](https://github.com/gravity-ui/markdown-editor/blob/cf5df476d9ab978f846beec296fed5ce8743e48a/src/extensions/yfm/ImgSize/utils.ts#L52) and [getImageSizeNew()](https://github.com/gravity-ui/markdown-editor/blob/cf5df476d9ab978f846beec296fed5ce8743e48a/src/extensions/yfm/ImgSize/utils.ts#L68), and ignores the `enableNewImageSizeCalculation` option.

makhnatkin: I think what needs to be done is:

1. Insert image immediately without dimensions (like regular images), saving position as `insertedPos = tr.selection.from - imageNode.nodeSize`

1. Load dimensions asynchronously via promise-based `loadImageFromUrl()`(similar to `loadImage()`

1. Update dimensions via `tr.setNodeMarkup()` using `getImageSizeNew()` with respect to `opts.enableNewImageSizeCalculation`

ReFFaT: Initially planned to use these methods and avoid two transactions, but they depend on flags:

needToSetDimensionsForUploadedImages: true
enableNewImageSizeCalculation: true
So decided to handle SVG separately from them.

Fixed the code - I have now returned to the default behavior, which depends on the flags, and added a default width, which is set if an image doesn't specify its own width.

image

src/extensions/yfm/ImgSize/utils.ts
```diff
@@ -34,6 +34,23 @@ export const createImageNode =
...
+                ? getImageSizeNew(image)
+                : getImageSize(image);
+
+            return imgType.create({
+                ...attrs,
+                width: image.width || DEFAULT_SVG_WIDTH,
+                ...sizes,
+            });
+        }
+
```
d3m1d0v: we can combine this with previous code:

```js
const isSvg = checkSvg(result.url) || file.type === 'image/svg+xml';

if (opts.needDimensions || isSvg) {
      try {
          const sizes = await loadImage(file).then(
              opts.enableNewImageSizeCalculation ? getImageSizeNew : getImageSize,
          );
          Object.assign(attrs, sizes);
      } catch (err) {
          globalLogger.error(err);
          logger.error({error: err});

          if (isSvg) attrs.width = String(DEFAULT_SVG_WIDTH);
      }
  }
```

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -91,6 +101,40 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
+                                                undefined,
+                                                {
+                                                    ...imageNode.attrs,
+                                                    width: img.width || DEFAULT_SVG_WIDTH,
+                                                    ...sizes,
+                                                },
+                                            );
+                                            view.dispatch(updateTr);
+
+                                            logger.event({
```
d3m1d0v, line 150: In this case, `logger.log` is preferred

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -91,6 +101,40 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
+
+                                            const updateTr = view.state.tr.setNodeMarkup(
+                                                insertedPos,
+                                                undefined,
+                                                {
+                                                    ...imageNode.attrs,
+                                                    width: img.width || DEFAULT_SVG_WIDTH,
+                                                    ...sizes,
+                                                },
+                                            );
```
d3m1d0v: plz use `tr.setNodeAttribute()` to set one attribute:
https://prosemirror.net/docs/ref/#transform.Transform.setNodeAttribute

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -91,6 +101,40 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
+                                    const insertedPos = tr.selection.from - imageNode.nodeSize;
+
+                                    loadImageFromUrl(imageUrl)
+                                        .then((img) => {
+                                            const sizes = opts?.enableNewImageSizeCalculation
+                                                ? getImageSizeNew(img)
+                                                : getImageSize(img);
+
+                                            const updateTr = view.state.tr.setNodeMarkup(
+                                                insertedPos,
```
d3m1d0v: position of image may have changed at this moment, so we need to track inserted image and get its current position

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -82,14 +89,81 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
                                 e.preventDefault();
 
+                                const isSvg = checkSvg(imageUrl);
+
+                                const trackingId = `img-${Math.random().toString(36).slice(2)}`;
+
                                 const imageNode = imageType(view.state.schema).create({
                                     src: imageUrl,
                                     alt: title,
+                                    ...(trackingId && {id: trackingId}),
```
d3m1d0v: `trackingId` always exists, it doesn't need to be checked

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -82,14 +89,81 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
                                 });
 
                                 const tr = view.state.tr.replaceSelectionWith(imageNode);
                                 view.dispatch(tr.scrollIntoView());
-                                logger.event({event: 'paste-url-as-image'});
+                                logger.log('paste-url-as-image');
```
d3m1d0v: `logger.log` -> `logger.event`

src/extensions/yfm/ImgSize/ImagePaste/index.ts
d3m1d0v: maybe this check should be `..(isSvg && {id: trackingId})` ?

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -82,14 +89,81 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
+                                            currentState.doc.descendants((node, pos) => {
+                                                if (
+                                                    isImageNode(node) &&
+                                                    node.attrs.id === trackingId
+                                                ) {
+                                                    targetPos = pos;
+                                                    return false;
+                                                }
+                                                return true;
+                                            });
```
d3m1d0v: move to helper function

```ts
function findImageNode(doc: Node, id: string): {node: Node; pos: number} | null;
```

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -82,14 +89,81 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
                                     alt: title,
+                                    ...(trackingId && {id: trackingId}),
                                 });
 
                                 const tr = view.state.tr.replaceSelectionWith(imageNode);
                                 view.dispatch(tr.scrollIntoView());
-                                logger.event({event: 'paste-url-as-image'});
+                                logger.log('paste-url-as-image');
+
+                                if (trackingId) {
```
d3m1d0v: `trackingId` always exist

src/extensions/yfm/ImgSize/ImagePaste/index.ts
```diff
@@ -82,14 +89,81 @@ export const ImagePaste: ExtensionAuto<ImagePasteOptions> = (builder, opts) => {
...
+                                    loadImageFromUrl(imageUrl)
+                                        .then((img) => {
+                                            const sizes: {
+                                                [ImgSizeAttr.Height]?: string;
+                                                [ImgSizeAttr.Width]?: string;
+                                            } =
+                                                opts?.enableNewImageSizeCalculation || isSvg
+                                                    ? getImageSizeNew(img)
+                                                    : getImageSize(img);
```
d3m1d0v: maybe a little more pretty?)

```js
loadImageFromUrl(file)
  .then(opts?.enableNewImageSizeCalculation || isSvg ? getImageSizeNew : getImageSize)
  .then(
    (sizes) => {
      // update image sizes
    },
    (err) => {
      // log error
    },
  );
```

src/extensions/yfm/ImgSize/ImagePaste/index.ts
ReFFaT: we've already decided that we'll assign an id to all images

src/extensions/yfm/ImgSize/ImagePaste/index.ts
ReFFaT: maybe like this? I think the method with the 2-parameter function is awesome

```
loadImageFromUrl(file)
  .then(opts?.enableNewImageSizeCalculation || isSvg ? getImageSizeNew : getImageSize)
  .then(
    (sizes) => {
      // update image sizes
    },
  )
  .catch(err => {
     // error log
  });
```

PR #882, opened 2025-11-21
feat(wysiwyg): add search and replace functionality to wysiwyg mode

d3m1d0v:
1. Updated styles for search matches in markup-mode
2. Updated appearance of search panel:
   - added compact view
   - also react components for search panel moved from `src/markup/codemirror/*` to `src/modules/search`
3. Added search-n-replace functionality to WYSIWYG-mode, same as in markup-mode
   - added Search behavior extension, uses `prosemirror-search` package
   - uses search components from `src/modules/search`
   - enabled by default, but can be disabled via props:
     ```js
     useMarkdownEditor({
       wysiwygConfig: {
         searchPanel: false,
       },
     });
     ```

### DIFF
--- a/tests/playwright/core/editor.ts
+++ b/tests/playwright/core/editor.ts
@@ -1,6 +1,7 @@
 import type {Expect, Locator, Page} from '@playwright/test';
 
 import type {DataTransferType, MarkdownEditorMode} from 'src';
+import {SearchQA} from 'src/modules/search';
 
 type YfmTableCellMenuType = 'row' | 'column';
 type YfmTableActionKind =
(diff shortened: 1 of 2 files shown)


src/modules/search/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './types';
+export * from './SearchCard';
+export * from './SearchPopup';
```
makhnatkin: I suggest putting all the editor's interface stuff in the modules folder. Let's move the search folder there.

src/extensions/behavior/Search/utils/connect-tracker.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+type Dispose = () => void;
+
+export function startTracking(elem: HTMLElement, options: TrackingOptions): Dispose {
+    const document = elem.ownerDocument;
+
+    let connected = document.contains(elem);
+
+    const observer = new MutationObserver(() => {
```
makhnatkin, line 13: It looks like we're using `startTracking` because the editor doesn't have `EventBus`. I suggest creating a task and linking to the issue here.

src/extensions/behavior/Search/utils/wrap-command.ts
```diff
@@ -0,0 +1,16 @@
+import type {Command, Transaction} from '#pm/state';
+
+export function wrapCommand(command: Command, wrapper: (tr: Transaction) => void): Command {
```
makhnatkin: It seems that `wrapCommand` can be used in many places and it can also be more versatile – accepting not just one callback, but an array of callbacks.

src/extensions/behavior/Search/utils/wrap-command.ts
```diff
@@ -0,0 +1,16 @@
...
+            dispatch
+                ? (tr) => {
+                      wrapper(tr);
+                      return dispatch(tr);
+                  }
+                : undefined,
+            view,
+        );
+    };
+}
```
makhnatkin, line 20: I suggest writing unit tests for simple utility functions.

src/extensions/behavior/Search/utils/connect-tracker.ts
d3m1d0v: yes, I created issue: https://github.com/gravity-ui/markdown-editor/issues/884

makhnatkin (approved):

PR #883, opened 2025-11-24
fix(Checkbox): handle checkbox without id attribute

makhnatkin:


src/extensions/yfm/Checkbox/CheckboxSpecs/schema.ts
```diff
@@ -46,11 +46,18 @@ export const getSchemaSpecs = (
...
-                    const checked = (node as HTMLInputElement).checked ? 'true' : null;
-                    const text = node.parentNode?.querySelector<HTMLLabelElement>(
-                        `label[for=${id}]`,
-                    )?.textContent;
+                    const element = node as HTMLElement;
+                    const id = element.id;
+                    const checked = (element as HTMLInputElement).checked ? 'true' : null;
+
+                    // find label by id if present, otherwise find next sibling label
+                    const labelById = element.parentNode?.querySelector<HTMLLabelElement>(
```
ogonkov: You can use labels property https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels

src/extensions/yfm/Checkbox/CheckboxSpecs/schema.ts
```diff
@@ -46,11 +46,18 @@ export const getSchemaSpecs = (
...
+                        `label[for="${id}"]`,
```
ogonkov: I guess this could be invalid selector after concatenation, i think it's worth to add try..catch, to handle possible exception

src/extensions/yfm/Checkbox/CheckboxSpecs/schema.ts
```diff
@@ -46,11 +46,18 @@ export const getSchemaSpecs = (
                 tag: 'input[type=checkbox]',
                 priority: 50,
                 getContent(node, schema) {
-                    const id = (node as HTMLElement).id;
-                    const checked = (node as HTMLInputElement).checked ? 'true' : null;
-                    const text = node.parentNode?.querySelector<HTMLLabelElement>(
-                        `label[for=${id}]`,
-                    )?.textContent;
+                    const element = node as HTMLElement;
```
ogonkov: You can add type guard to avoid casting below


```suggestion
                    const element = node instanceof HTMLInputElement ? node : undefined;
```

src/extensions/yfm/Checkbox/CheckboxSpecs/schema.ts
```diff
@@ -46,7 +46,7 @@ export const getSchemaSpecs = (
                 tag: 'input[type=checkbox]',
                 priority: 50,
                 getContent(node, schema) {
-                    const element = node as HTMLElement;
+                    const element = node instanceof HTMLInputElement ? node : undefined;
                     const id = element.id;
                     const checked = (element as HTMLInputElement).checked ? 'true' : null;
```
ogonkov: Casting here should be unnecessary

ogonkov (approved):

PR #888, opened 2025-12-05
feat(Math): parse TeX formulas when pasting from code editor and paste them as math block

ReFFaT:
added a Math plugin that processes text with a mod on the input that contains formulas

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22265,9 +22265,9 @@
       }
     },
     "node_modules/vite": {
-      "version": "5.4.19",
-      "resolved": "https://registry.npmjs.org/vite/-/vite-5.4.19.tgz",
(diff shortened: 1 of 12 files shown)


src/extensions/markdown/CodeBlock/handle-paste.ts
```diff
@@ -44,6 +45,10 @@ function getCodeData(data: DataTransfer): null | {editor: string; mode?: string;
         if (isVSCode(data)) {
             editor = 'vscode';
             mode = tryParseVSCodeData(data)?.mode;
+
+            if (isLatexMode(mode)) {
```
makhnatkin: `CodeBlock` is now dependent on `Math`.  This violates architectural principles.
I suggest solving this with the help of priorities, if `High` does not help, then increase

ReFFaT: fixed)

PR #899, opened 2025-12-14
feat(CodeBlock): updated code paste

ReFFaT:
The code insertion functionality has been updated: the system now correctly recognizes code pasted from WebStorm and similar IDEs, as well as from certain other sources, including our AI. Inline code support has also been added—if the user inserts a single line, it is formatted as inline code rather than a code block.

### DIFF
--- a/src/extensions/markdown/CodeBlock/handle-paste.ts
+++ b/src/extensions/markdown/CodeBlock/handle-paste.ts
@@ -12,6 +12,7 @@ export type CodePasteData = {
     editor: string;
     value: string;
     inline: boolean;
+    mode?: string;
 };
 
 type InsertCodeParams = {
@@ -35,7 +36,15 @@ export const handlePaste: NonNullable<EditorProps['handlePaste']> = (view, e) =>
     const $from = selection.$from;
     const inCodeBlock = Boolean($from.parent.type.spec.code);
 
-    logPasteEvent(state, code, data);
+    getLoggerFromState(state).event({
+        domEvent: 'paste',
+        event: 'paste-from-code-editor',
+        editor: code.editor,
+        editorMode: code.mode,
+        empty: !code.value,
+        inline: code.inline,
+        dataTypes: Array.from(data.types),
+    });
 
     if (!code.value) {
         return false;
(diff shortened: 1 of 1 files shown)


src/extensions/markdown/CodeBlock/handle-paste.ts
```diff
@@ -1,52 +1,147 @@
...
+
+export function isInlineCode(text: string): boolean {
+    return !text.includes('\n');
+}
+
+export function isInlineCodeFromHtml(html: string, text: string): boolean {
+    return html.includes('<code') && !html.includes('<pre') && isInlineCode(text);
+}
+
+function dedentIfMultiline(value: string): string {
```
makhnatkin: `dedentIfMultiline` looks redundant, `dd` already effectively does the same thing

src/extensions/markdown/CodeBlock/handle-paste.ts
```diff
@@ -1,52 +1,147 @@
...
+    const codeBlockNode = nodeType.create(null, textNode);
+    tr.replaceSelectionWith(codeBlockNode);
+}
+
+export function getCodeData(data: DataTransfer): CodePasteData | null {
+    const text = data.getData(DataTransferType.Text);
+    if (!text) return null;
+
+    const vscodeData = parseVSCodeData(data, text);
+    if (vscodeData) return vscodeData;
```
makhnatkin: `getCodeData()` currently relies on `parse*Data()` functions that both detect the source (VSCode/HTML) and process it. This creates a "detect → parse → check result" flow, which is less clear and harder to extend

I suppose:
1. Detect the paste source once in `getCodeData()`
1. Keep processing logic in dedicated `process*` functions
1. Easier to read and extend with new sources

Example:

```
export function getCodeData(data: DataTransfer): CodePasteData | null {
  const text = data.getData(DataTransferType.Text);
  if (!text) return null;

  // Detect and process VSCode
  if (isVSCode(data)) {
    return processVSCodePaste(data, text);
  }

  // Detect and process HTML
  const html = data.getData('text/html') || '';
  if (html && (html.includes('<pre') || html.includes('<code'))) {
    return processHtmlPaste(data, text);
  }

  return null;
}

function processVSCodePaste(data: DataTransfer, text: string): CodePasteData {
  const vsCodeData = tryParseVSCodeData(data);
  return {
    editor: 'vscode',
    value: dedentIfMultiline(text),
    inline: isInlineCode(text),
    mode: vsCodeData?.mode,
  };
}

function processHtmlPaste(data: DataTransfer, text: string): CodePasteData {
  const html = data.getData('text/html') || '';
  const inline = isInlineCodeFromHtml(html, text);
  return {
    editor: 'code-editor',
    value: inline ? text : dedentIfMultiline(text),
    inline,
  };
}
```

* Source detection happens once
* Clear separation: detect in getCodeData(), process in process*() functions
* Easy to add new paste sources

You could make this more declarative using a hashmap approach, but I'd keep it simple for now with just two sources

src/extensions/markdown/CodeBlock/handle-paste.ts
```diff
@@ -1,52 +1,147 @@
...
+        to: selection.to,
+        inCodeBlock,
+    });
+
+    view.dispatch(tr.scrollIntoView());
+    e.preventDefault();
+    return true;
+};
+
+function logPasteEvent(
```
makhnatkin: `logPasteEvent` is used only once, so inlining it might be simpler.

Also, the logging lost the `editorMode` field, is it ok?

makhnatkin (approved):

PR #900, opened 2025-12-15
dev: migrate to pnpm

d3m1d0v:


.github/workflows/release.yml
d3m1d0v: TODO: update release workflow

package.json
```diff
@@ -245,10 +262,10 @@
     "@playwright/test": "1.49.0",
     "@storybook/addon-docs": "10.1.10",
     "@storybook/addon-webpack5-compiler-babel": "4.0.0",
-    "@storybook/cli": "^10.1.10",
+    "@storybook/cli": "10.1.10",
     "@storybook/preset-scss": "1.0.3",
     "@storybook/react": "10.1.10",
-    "@storybook/react-webpack5": "^10.1.10",
+    "@storybook/react-webpack5": "10.1.10",
```
makhnatkin, line 268: Is this version pinning on purpose, or is it just from the migration?

.npmrc
```diff
@@ -1,2 +1,4 @@
 registry=https://registry.npmjs.org/
 save-exact=true
+update-notifier=false
+auto-install-peers=true
```
makhnatkin, line 4: It can hide missing peer dependency issues and make installs less predictable. Why is that better here than keeping peer deps explicit?

src/core/markdown/MarkdownParser.ts
```diff
@@ -1,10 +1,9 @@
-import type {Match} from 'linkify-it'; // eslint-disable-line import/no-extraneous-dependencies
 import type MarkdownIt from 'markdown-it';
 import type Token from 'markdown-it/lib/token';
 import {Mark, type MarkType, type Node, type NodeType, type Schema} from 'prosemirror-model';
 
 import {type Logger2, logger} from '../../logger';
-import type {Parser, ParserToken} from '../types/parser';
+import type {LinkMatch, Parser, ParserToken} from '../types/parser';
```
makhnatkin, line 6: This looks unrelated to the pnpm migrationю Why did this have to be updated as part of switching to pnpm?

d3m1d0v: I didn't add `linkify-it` to package dependencies, so I had to replace import of `Match` with import of `LinkMatch`

package.json
d3m1d0v: versions of @storybook/* packages must be in sync

.npmrc
d3m1d0v: This setting only affects development, and packages won't appear in root of `node_modules` and won't be able to be imported. Therefore, this won't cause any problems

PR #917, opened 2026-01-12
feat(view): add support of yfm-files to `useFilesGallery()` hook

d3m1d0v:
Added support for YFM files to `useFilesGallery()` hook: getting the file name and type from element attributes


src/view/hooks/useFilesGallery/helpers.ts
```diff
@@ -0,0 +1,34 @@
+import {YfmFileAttrs, YfmFileClassname} from './constants';
+
```
kseniya57, line 2: Shouldn't this file be called utils.ts? I see that this name is more commonly used in the project

PR #923, opened 2026-01-14
feat(markup): support codemirror tooltips config

saxumcordis:
We have problems with tooltips and editor's `overflow: hidden`. 
These changes allow to control tooltips parent


src/markup/codemirror/gravity.ts
```diff
@@ -24,7 +24,7 @@ export const gravityHighlightStyle = HighlightStyle.define(
 );
 
 export const gravityTheme = EditorView.baseTheme({
-    '&': {
+    '&.cm-editor': {
```
saxumcordis: image

tooltip with custom parent is wrapped with `& class` which has overflow hidden :) 
It should not have this one to fix initial problem.. so this fixes it

saxumcordis: maybe my changes are not needed and we should remove `overflow: hidden` at all..

saxumcordis: decided to make remove overflow completely: https://github.com/gravity-ui/markdown-editor/pull/927

PR #928, opened 2026-01-21
ci: add manual publishing workflow

d3m1d0v:
- warning (instead of error) when package manager mismatch
- add manual publishing workflow
- disable auto publishing from release workflow
- remove prerelease workflow and use publish workflow instead


.github/workflows/release.yml
```diff
@@ -51,10 +51,4 @@ jobs:
...
         if: steps.release.outputs.release_created == 'true'
         shell: bash
-        run: |
-          echo "steps.release.outputs.release_created: ${{ steps.release.outputs.release_created }}"
-          pnpm config set '//registry.npmjs.org/:_authToken' "${NODE_AUTH_TOKEN}"
-          pnpm publish --tag ${{ github.ref_name != 'main' && 'untagged' || 'latest' }} --access public --report-summary
-          cat pnpm-publish-summary.json
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.GRAVITY_UI_BOT_NPM_TOKEN }}
+        run: echo "Please run publish workflow manually"
```
makhnatkin, line 54: Manual publish is fine for now; let’s add auto-release later to reduce human error.

PR #937, opened 2026-01-23
feat(search): make search panel semi-transparent when overlapping active search match

d3m1d0v:


src/modules/search/components/hooks/useObserveIntersection.ts
```diff
@@ -0,0 +1,71 @@
...
+    floatingRef,
+    container,
+    selector,
+    observerOptions,
+}: UseObserveIntersectionProps): UseObserveIntersectionReturn {
+    const [intersection, setIntersection] = useState<boolean>(false);
+    const intersectionRef = useLatest<boolean>(intersection);
+
+    const updateFunction = useCallback(() => {
+        const floaing = floatingRef.current;
```
makhnatkin: typo: floaing -> floa**t**ing

PR #945, opened 2026-02-03
feat(Clipboard): remove empty blocks at edges of selected content when copying

d3m1d0v:


src/extensions/behavior/Clipboard/trim-selection.ts
```diff
@@ -0,0 +1,83 @@
...
+    const to = trimTableEdgeCell($to, 'backward');
+    if ($from.pos === from && $to.pos === to) return sel;
+    if (from > to) return TextSelection.create($from.doc, to, to);
+    return TextSelection.create($from.doc, from, to);
+}
+
+/** @internal */
+export function trimTableEdgeCell($pos: ResolvedPos, dir: 'backward' | 'forward'): number {
+    let depth = $pos.depth;
+    let isAtEdgeOfInlineBlocks = true;
```
makhnatkin: nit
The variable doesn’t check “inline”; it checks the edge along the document tree. Let’s name it `isAtContentEdge` or `isAtTextblockEdge`.

src/extensions/behavior/Clipboard/table.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    {
+        const node = $pos.node(depth);
+        const parentNode = $pos.node(depth - 1);
+
+        const isInTable = node.type.spec.tableRole || parentNode.type.spec.tableRole;
+        if (!isInTable || !node.isTextblock) {
+            return $pos.pos;
+        }
```
makhnatkin: As I understand code correctly we only detect “in table” when the textblock or its direct parent has `tableRole`. If a cell content is wrapped (e.g. `tableCell` → `blockquote`/`list` → `p`), `isInTable` becomes `false` and we return early, so edge navigation/trim doesn’t work. We should walk up to the nearest ancestor with `tableRole` (cell) and then check whether the selection is at the edge of that cell (possibly via wrappers).

Let's add tests for 
* `yfmTd(blockquote(p('text'))):`
* `yfmTd(list(li(p('text')))):`

src/extensions/behavior/Clipboard/trim-selection.ts
```diff
@@ -0,0 +1,83 @@
...
+        const node = $pos.node(depth);
+        const parentNode = $pos.node(depth - 1);
+
+        if (!node.type.spec.tableRole || !parentNode.type.spec.tableRole) break;
+
+        const edgeNode = dir === 'forward' ? parentNode.lastChild : parentNode.firstChild;
+        if (node !== edgeNode) break;
+
+        depth--;
+    }
```
makhnatkin, line 55: nit 
If there is another table inside `tableCell`, `tableRole` will start to appear at several levels.

makhnatkin (approved):

makhnatkin (approved):

src/extensions/behavior/Clipboard/table.ts
d3m1d0v: trimming now works for all blocks, not just tableCell

src/extensions/behavior/Clipboard/trim-selection.ts
d3m1d0v: trimming now works for all blocks, not just tableCell

PR #946, opened 2026-02-04
fix(CodeBlock): handle pasting only from code-editors known to us

ReFFaT:
### Problem
Pasting HTML with code + other elements (tables, links) inserted everything as code block.

### Solution
Check if content is pasted from a code editor, and only then paste content as code.
Known code editors: vscode and jetbrains

### DIFF
--- a/src/extensions/markdown/CodeBlock/handle-paste.ts
+++ b/src/extensions/markdown/CodeBlock/handle-paste.ts
@@ -4,7 +4,7 @@ import dd from 'ts-dedent';
 
 import {getLoggerFromState} from '#core';
 import type {EditorProps} from '#pm/view';
-import {DataTransferType, isVSCode, tryParseVSCodeData} from 'src/utils/clipboard';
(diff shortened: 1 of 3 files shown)


src/extensions/markdown/CodeBlock/handle-paste.ts
```diff
@@ -124,6 +129,24 @@ function processHtmlPaste(data: DataTransfer, text: string): CodePasteData {
...
+    const codeElements = div.querySelectorAll('pre, code');
+    if (codeElements.length === 0) {
+        return false;
+    }
+
+    const clone = div.cloneNode(true) as HTMLElement;
+    clone.querySelectorAll('pre, code').forEach((el) => el.remove());
+
+    const remainingText = clone.textContent?.trim() || '';
+    return remainingText.length === 0;
```
d3m1d0v: inserted HTML may contain images, iframes, and other blocks that may not contain text

PR #949, opened 2026-02-10
feat(CodeBlock): improvements in code block

d3m1d0v:
Changes in **CodeBlock**:
- Text wrapping is disabled by default.
- Floating toolbar for code block has been redesigned:
  - Move _PlainText_ to first position in the language select list.
  - Replace the "line numbers" checkbox with a toggle button
  - Add a copy button
  - Change the remove button theme to danger
- Add button to the floating toolbar to toggle text wrapping - it will only affect the current code block and only during current editing session for current user

**Toolbar**:
- add danger theme to single button

Also added a new example story where you can configure a code block.

image


demo/stories/examples/code-block/CodeBlock.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import {MarkdownEditorView, useMarkdownEditor} from 'src/index';
+
+import {PlaygroundLayout} from '../../../components/PlaygroundLayout';
+
+import {markup} from './markup';
+
+export type YfmTableDnDDemoProps = {
+    lineWrapping: 'disabled' | 'enabled';
+    lineNumbers: 'disabled' | 'enabled' | 'by-default';
```
makhnatkin: In my opinion, `by-default` looks unconventional as an enum value. Maybe rename it to `default` or `auto`?

src/extensions/markdown/CodeBlock/CodeBlockHighlight/CodeBlockNodeView.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        if (this._lang) {
+            this.dom.setAttribute(CodeBlockNodeAttr.Lang, this._lang);
+            this._codeElem.classList.add(this._lang);
+        }
+
+        this.contentDOM = document.createElement('div');
+
+        const hasLineWrapping = hasLineWrappingDecoration(decorations);
+        if (this._opts.lineNumbers?.enabled && hasLineWrapping) {
```
makhnatkin: I’m not fully sure why `hasLineWrapping` is needed here, and why it isn’t needed in `update()` for `_applyLineNumbersResult`: 

```
if (this._opts.lineNumbers?.enabled) {
            this._applyLineNumbersResult = applyLineNumbers(
...
```

Could you clarify the reasoning?

demo/stories/examples/code-block/CodeBlock.tsx
d3m1d0v: fixed

src/extensions/markdown/CodeBlock/CodeBlockHighlight/CodeBlockNodeView.ts
d3m1d0v: fixed

PR #962, opened 2026-02-26
fix: mermaid edit theme dependency

Radomir-Drukh:
Right now mermaid is only rendered in "forest" theme that does not look good in dark theme. This PR fixes this and adds a dependency to track current application theme

### DIFF
--- a/demo/defaults/content.ts
+++ b/demo/defaults/content.ts
@@ -273,14 +273,6 @@ And ##monospace## can be **##com##**##bined\\*##
 
 {% endnote %}
 
-Small mermaid example: 
-
-\`\`\`mermaid
-sequenceDiagram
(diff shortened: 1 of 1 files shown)


demo/defaults/content.ts
```diff
@@ -273,6 +273,14 @@ And ##monospace## can be **##com##**##bined\\*##
...
 {% endnote %}
 
+Small mermaid example: 
+
+\`\`\`mermaid
+sequenceDiagram
+	Alice->>Bob: Hi Bob
+	Bob->>Alice: Hi Alice
+\`\`\`
+
```
d3m1d0v: Let's remove diagram from default demo content. A separate story and visual tests are sufficient for mermaid diagram.

PR #963, opened 2026-02-26
ci: add workflow to publish static docs

makhnatkin:


.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,65 @@
...
+          path: ./docs-dist
+
+  deploy:
+    name: Deploy to GitHub Pages
+    needs: build
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      # The deploy action outputs the final Pages URL here.
+      url: ${{ steps.deployment.outputs.page_url }}
```
makhnatkin, line 60: https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages#deploying-github-pages-artifacts

PR #979, opened 2026-03-10
fix(view): do not render mermaid inside wysiwyg editor

Radomir-Drukh:


packages/editor/src/view/hocs/withMermaid/index.tsx
```diff
@@ -28,7 +28,12 @@ export function withMermaid(opts: WithMermaidOptions) {
             useMermaidRuntime(meta, opts.runtime);
 
             useEffect(() => {
-                renderMermaid(mermaidConfig);
+                // do not render mermaid diagrams inside the editor component
+                // because they are handled differently (no data-content attribute)
+                // and the editor is managing their render itself
+                renderMermaid(mermaidConfig, {
+                    querySelector: '.mermaid:not(.g-md-editor-component .mermaid)',
```
d3m1d0v: invalid selector

PR #990, opened 2026-03-14
feat(core): allow independent registration and override of node/mark specs in extensions

d3m1d0v:
## Description

`addNode`/`addMark` bundle schema spec, parser spec, and serializer spec into a single callback, making it impossible for one extension to incrementally modify another — the only option was full replacement.

This PR adds granular registration and override methods to `ExtensionBuilder`:

**New add methods:**

- `addNodeSpec`, `addMarkSpec` — register ProseMirror schema specs independently
- `addMarkdownTokenParserSpec` — register markdown-it token → PM entity mapping
- `addNodeSerializerSpec`, `addMarkSerializerSpec` — register serializer specs

**New override methods:**

- `overrideNodeSpec`, `overrideMarkSpec` — modify previously registered schema specs
- `overrideMarkdownTokenParserSpec` — modify parser token specs (works with both `addNode` and granular API)
- `overrideNodeSerializerSpec`, `overrideMarkSerializerSpec` — modify serializer specs

`addNode` and `addMark` methods are marked as `@deprecated`.

### DIFF
--- a/packages/editor/src/core/ExtensionBuilder.test.ts
+++ b/packages/editor/src/core/ExtensionBuilder.test.ts
@@ -211,4 +211,866 @@ describe('ExtensionBuilder', () => {
 
         expect(fn).toThrow(Error);
     });
+
+    describe('granular add methods', () => {
+        it('should add node via granular methods', () => {
+            const toMd = jest.fn();
+            const nodes = new ExtensionBuilder(logger)
+                .addNodeSpec('myNode', () => ({group: 'block'}))
+                .addMarkdownTokenParserSpec('my_node', () => ({
+                    name: 'myNode',
+                    type: 'block',
+                }))
+                .addNodeSerializerSpec('myNode', () => toMd)
+                .build()
+                .nodes();
+
+            expect(nodes.size).toBe(1);
+            const nodeSpec = nodes.get('myNode');
+            expect(nodeSpec).toBeTruthy();
+            expect(nodeSpec!.spec.group).toBe('block');
+            expect(nodeSpec!.fromMd.tokenName).toBe('my_node');
+            expect(nodeSpec!.fromMd.tokenSpec.name).toBe('myNode');
+            expect(nodeSpec!.fromMd.tokenSpec.type).toBe('block');
+            expect(nodeSpec!.toMd).toBe(toMd);
+        });
+
+        it('should add mark via granular methods', () => {
+            const marks = new ExtensionBuilder(logger)
+                .addMarkSpec('myMark', () => ({
+                    parseDOM: [{tag: 'em'}],
+                    toDOM() {
+                        return ['em'];
+                    },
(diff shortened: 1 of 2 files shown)


packages/editor/src/core/ExtensionBuilder.test.ts
```diff
@@ -211,4 +211,526 @@ describe('ExtensionBuilder', () => {
 
         expect(fn).toThrow(Error);
     });
+
+    describe('granular add methods', () => {
+        it('should add node via granular methods', () => {
```
makhnatkin, line 216:
```ts
it('should preserve addNode full spec when addMarkdownTokenParserSpec uses same tokenName', () => {
    // Bug scenario: tokenName === entityName in step 1b of build()
    // Step 1 adds full spec under 'code_block'.
    // Step 1b finds parserSpecsByEntity['code_block'] and calls map.addToEnd('code_block', parserOnlyEntry),
    // which overwrites the full spec with { spec: {}, toMd: () => { throw } }.
    const realToMd = jest.fn();

    const nodes = new ExtensionBuilder(logger)
        .addNode('code_block', () => ({
            spec: {group: 'block', code: true},
            fromMd: {tokenSpec: {name: 'code_block', type: 'block', noCloseToken: true}},
            toMd: realToMd,
        }))
        .addMarkdownTokenParserSpec('code_block', () => ({
            // tokenName === entityName: 'code_block' → 'code_block'
            name: 'code_block',
            type: 'block',
            noCloseToken: true,
        }))
        .build()
        .nodes();

    // Full addNode spec must survive — not be replaced by parser-only entry
    expect(nodes.size).toBe(1);

    const codeBlock = nodes.get('code_block');
    expect(codeBlock).toBeTruthy();
    // Spec must be the real one, not {} from buildParserOnlyNodeEntry
    expect(codeBlock!.spec.group).toBe('block');
    expect(codeBlock!.spec.code).toBe(true);
    // toMd must be the real function, not the throwing stub
    expect(codeBlock!.toMd).toBe(realToMd);
    expect(() => (codeBlock!.toMd as Function)()).not.toThrow();
});
```

packages/editor/src/core/ExtensionBuilder.ts
```diff
@@ -110,6 +189,10 @@ export class ExtensionBuilder {
         return this;
     }
 
+    /**
+     * @deprecated Will be removed in the next major version.
+     * Use addNodeSpec() + addMarkdownTokenParserSpec() + addNodeSerializerSpec() instead.
+     */
     addNode(name: string, cb: AddPmNodeCallback): this {
```
makhnatkin, line 196: Should we add the same checks like:
```ts

        if (this.#rawNodeSpecs[name]) {
            throw new Error(`Node spec with name "${name}" already registered via addNodeSpec`);
        }
```
?

PR #1029, opened 2026-03-25
perf(CodeBlock): reduce number of code block tooltip re-renders

d3m1d0v:
## Summary

Reduce unnecessary re-renders in the code block tooltip.

- Removed a `window.dispatchEvent(new CustomEvent('scroll'))` hack that was triggering re-renders globally on each tooltip update
- Added `noRerenderOnUpdate` flag to `ToolbarReactNode` and `ToolbarReactNodeFn` to opt out of forced re-renders for static toolbar items (e.g. copy button)
- Extracted inline toolbar into `CodeBlockToolbar` and `CodeLangSelect` to enable component-level memoization

### DIFF
--- a/packages/editor/src/toolbar/ToolbarGroup.tsx
+++ b/packages/editor/src/toolbar/ToolbarGroup.tsx
@@ -1,3 +1,5 @@
+import {Fragment} from 'react';
+
 import {cn} from '../classname';
 
 import {ToolbarButton} from './ToolbarButton';
@@ -86,7 +88,7 @@ export function ToolbarButtonGroup<E>({
                         />
                     );
                     return item.noRerenderOnUpdate ? (
-                        renderFn()
+                        <Fragment key={item.id}>{renderFn()}</Fragment>
                     ) : (
                         <ToolbarUpdateOnRerender key={item.id} content={renderFn} />
                     );
@@ -94,7 +96,7 @@ export function ToolbarButtonGroup<E>({
 
                 if (item.type === ToolbarDataType.ReactNode) {
                     return item.noRerenderOnUpdate ? (
-                        item.content
+                        <Fragment key={item.id}>{item.content}</Fragment>
                     ) : (
                         <ToolbarUpdateOnRerender key={item.id} content={() => item.content} />
                     );
@@ -102,7 +104,7 @@ export function ToolbarButtonGroup<E>({
 
                 if (item.type === ToolbarDataType.ReactNodeFn) {
                     return item.noRerenderOnUpdate ? (
-                        item.content(editor)
+                        <Fragment key={item.id}>{item.content(editor)}</Fragment>
                     ) : (
                         <ToolbarUpdateOnRerender
                             key={item.id}


packages/editor/src/toolbar/types.ts
```diff
@@ -120,6 +120,7 @@ export type ToolbarReactNodeData = {
     type: ToolbarDataType.ReactNode;
     width: number;
     content: React.ReactNode;
+    noRerenderOnUpdate?: boolean;
```
makhnatkin, line 123: Am I right that the current approach solves the problem by filtering a broad update signal: we keep a wide update mechanism, but manually disable rerender for some toolbar items? If so, how safe is this in the long term, and should we invert the model so that items do not listen to updates by default, and only required items subscribe to specific update types?

makhnatkin: Do we have a clear list of which toolbar items depend on attrs, which depend on decorations, and which should not react to updates at all?

Maybe we should consider the opposite model: toolbar items do not listen to updates by default, and only required ones subscribe to specific update types. As a first step, three signal types may already be enough: `stateUpdate`, `layoutUpdate`, and `selectionUpdate`. This could reduce overall noise and make item dependencies more explicit.

### 1. `stateUpdate`
Use this when the underlying node state or attrs changed.

Examples:
- code block language changed
- line numbers flag changed
- another node attribute changed

This type means: **the data of the block changed**.

### 2. `layoutUpdate`
Use this when the visual layout or derived visual state changed.

Examples:
- wrapping changed
- decorations changed
- tooltip/popup geometry may need refresh

This type means: **the visual presentation changed**, even if the raw node attrs did not.

### 3. `selectionUpdate`
Use this when the current selection or active node context changed.

Examples:
- user moved selection to another node
- toolbar needs to reflect a different active block
- item state depends on which node is currently selected

This type means: **the editing context changed**.

### Current model
- one global `update`
- many items listen to it
- some items are filtered out with `noRerenderOnUpdate`

### Proposed model
- an item listens to **nothing by default**
- if it needs updates, it explicitly declares which update types it listens to

This makes dependencies clearer.


### Example possible type definition:

```ts
type ToolbarUpdateType = 'stateUpdate' | 'layoutUpdate' | 'selectionUpdate';

type ToolbarReactComponentData<E> = {
  id: string;
  type: ToolbarDataType.ReactComponent;
  component: React.ComponentType<ToolbarBaseProps<E>>;
  props?: object;
  updateOn?: ToolbarUpdateType[];
};
```

Meaning:
	•	if `updateOn` is not set, the item does not listen to updates
	•	if `updateOn: ['stateUpdate']`, the item listens only to `stateUpdate`

### Example subscription:

```ts
// now
context.eventBus.on('update', rerender);

// proposal
for (const eventName of updateOn) {
  context.eventBus.on(eventName, rerender);
}

```

### Example code block toolbar
```ts
// copy button
{
  id: 'code-block-copy',
  type: ToolbarDataType.ReactNodeFn,
  content: () => <ClipboardButton text={copyText} />,
}

// language select
{
  id: 'code-block-type',
  type: ToolbarDataType.ReactComponent,
  component: CodeLangSelectItem,
  props: {...},
  updateOn: ['stateUpdate'],
}

// wrapping toggle
{
  id: 'code-block-wrapping',
  type: ToolbarDataType.SingleButton,
  ...,
  updateOn: ['layoutUpdate'],
}

```

packages/editor/src/toolbar/ToolbarGroup.tsx
```diff
@@ -93,11 +93,17 @@ export function ToolbarButtonGroup<E>({
...
+                        item.content
+                    ) : (
+                        <ToolbarUpdateOnRerender key={item.id} content={() => item.content} />
+                    );
                 }
 
                 if (item.type === ToolbarDataType.ReactNodeFn) {
-                    return (
+                    return item.noRerenderOnUpdate ? (
+                        item.content(editor)
```
makhnatkin: With `noRerenderOnUpdate` we return React nodes without key. Could this cause warnings or reconciliation issues?

We can wrap content in `Fragment` with `key={item.id}` to keep stable list identity.

packages/editor/src/extensions/markdown/CodeBlock/CodeBlockHighlight/TooltipPlugin/CodeBlockToolbar.tsx
```diff
@@ -0,0 +1,168 @@
...
+    rerenderTooltip,
+}: CodeBlockToolbarProps) {
+    const posRef = useLatest(pos);
+    const nodeRef = useLatest(node);
+
+    const onFocus = useCallback(() => {
+        editorView.focus();
+    }, [editorView]);
+
+    const toolbarData = useMemo<ToolbarData<EditorView>>(() => {
```
makhnatkin, line 58: I noticed that inside `toolbarData` we create new inline functions (`component`, `content`, `exec`) on each recompute.

Even if the data is the same, these functions have new references, which may reduce the real benefit of memoization.

Would it make sense to stabilize some of these callbacks with `useCallback`, or in some cases switch to `ReactComponent + props` to avoid inline functions or useCallback?

```ts
// before
const toolbarData = useMemo(() => {
  return [
    {
      id: 'copy',
      type: ToolbarDataType.ReactNodeFn,
      content: () => <ClipboardButton text={nodeRef.current.textContent} />,
    },
  ];
}, [nodeRef]);

// Example ReactComponent + props:
const CopyButtonItem = memo(function CopyButtonItem({text}) {
  return <ClipboardButton text={text} />;
});

const toolbarData = useMemo(() => {
  return [
    {
      id: 'copy',
      type: ToolbarDataType.ReactComponent,
      component: CopyButtonItem,
      props: {
        text: nodeRef.current.textContent,
      },
    },
  ];
}, [nodeRef]);

// Example useCallback:
const copyText = useCallback(() => {
  return nodeRef.current.textContent;
}, [nodeRef]);

const renderCopy = useCallback(() => {
  return <ClipboardButton text={copyText()} />;
}, [copyText]);

const toolbarData = useMemo(() => {
  return [
    {
      id: 'copy',
      type: ToolbarDataType.ReactNodeFn,
      content: renderCopy,
    },
  ];
}, [renderCopy]);


```

packages/editor/src/plugins/BaseTooltip/index.tsx
```diff
@@ -232,8 +232,6 @@ export class BaseTooltipPluginView implements PluginView {
 
     protected renderContent(currentNode: BaseTooltipNode): React.ReactNode {
         if (!this.content) return null;
-        // hack for popup rerender
-        window.dispatchEvent(new CustomEvent('scroll'));
```
makhnatkin, line 236: It's good that this was removed 🚀

packages/editor/src/extensions/markdown/CodeBlock/CodeBlockHighlight/TooltipPlugin/CodeBlockToolbar.tsx
d3m1d0v: No, because one call to `useMemo()` is better than many calls to `useMemo()` and `useCallback()`

PR #1037, opened 2026-03-30
perf(SelectionContext): reduce number of selection context tooltip re-renders

d3m1d0v:
Reduce unnecessary re-renders in the selection-context tooltip by memoizing toolbar component and its config

### DIFF
--- a/packages/editor/src/extensions/behavior/SelectionContext/TextSelectionTooltip.tsx
+++ b/packages/editor/src/extensions/behavior/SelectionContext/TextSelectionTooltip.tsx
@@ -1,0 +1,93 @@
+import {useEffect, useMemo, useState} from 'react';
+
+import {Popup, type PopupPlacement, type PopupProps, sp} from '@gravity-ui/uikit';
+
+import type {ActionStorage} from '#core';
+import type {EditorView} from '#pm/view';
+import {isFunction} from 'src/lodash';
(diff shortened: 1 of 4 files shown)


packages/editor/src/extensions/behavior/SelectionContext/TextSelectionTooltip.tsx
```diff
@@ -0,0 +1,93 @@
...
+        editorView,
+    }) {
+        const [conditionKey, setConditionKey] = useState(() =>
+            calcConditionKey(config, editor, editorView),
+        );
+
+        useEffect(() => {
+            const newKey = calcConditionKey(config, editor, editorView);
+            if (conditionKey !== newKey) setConditionKey(newKey);
+        });
```
makhnatkin, line 47: `useEffect` without deps runs on every render. When key changes, `setConditionKey` adds one extra render. Maybe you can just compute it inline:                                                                                                         

```ts                                                                                                                                          
  const conditionKey = calcConditionKey(config, editor, editorView);                                                                      
```   

 `useMemo` still works the same — recalculates only when key changes.

d3m1d0v: I specifically moved `сonditionKey` calculation into `useEffect()` to make render flow as easy as possible, since I assume that it will not change with most re-renders.

PR #1041, opened 2026-04-01
fix(Lists): improve nested list collapse for deep structures

makhnatkin:
Fixes https://github.com/gravity-ui/markdown-editor/issues/1042

### Improvements
- **O(N) single-pass algorithm**: replaced iterative O(K*N) approach with a single tree traversal that collects all redundant nesting in one pass
- **Recursive flattening**: handles deeply nested structures (up to 100 levels) with proper depth guards
- **Correct position mapping**: applies replacements in reverse order to preserve document positions, uses `tr.mapping` for accurate cursor placement

### DIFF
--- a/packages/editor/src/extensions/markdown/Lists/plugins/CollapseListsPlugin.ts
+++ b/packages/editor/src/extensions/markdown/Lists/plugins/CollapseListsPlugin.ts
@@ -145,4 +145,3 @@ function collapseListItemContent(itemNode: Node, schema: Schema, depth: number):
 
     return result;
 }
-


packages/editor/src/extensions/markdown/Lists/plugins/CollapseListsPlugin.ts
```diff
@@ -1,96 +1,172 @@
...
+
+            // collapse of parasitic nested lists
+            const lastCollapsePos = collapseAllNestedListItems(tr);
+
+            // restore cursor position
+            if (lastCollapsePos !== null) {
+                const cursorPos = findClosestTextNodePos(tr.doc, lastCollapsePos);
+                if (cursorPos !== null) {
+                    tr.setSelection(TextSelection.create(tr.doc, cursorPos));
+                }
```
d3m1d0v: u can use `Selection.findFrom()` or `Selection.near()` instead of custom logic

packages/editor/src/extensions/markdown/Lists/plugins/CollapseListsPlugin.ts
```diff
@@ -1,96 +1,172 @@
...
+    // skipUntil prevents re-processing children that were already collapsed
+    // by the recursive call in collapseListItemContent
+    let skipUntil = -1;
+
+    tr.doc.descendants((node, pos) => {
+        if (pos < skipUntil) return false;
+
+        if (!isListItemNode(node)) {
+            return true;
+        }
```
d3m1d0v, line 76: I suggest not to dive into text blocks – there definitely can’t be lists there.

PR #1043, opened 2026-04-01
feat(Link): do not insert url protocol for links from buffer in visible text

saxumcordis:
## Summary by Sourcery

Adjust link paste handling to distinguish between link href and displayed label when pasting text into the editor.

Bug Fixes:
- Prevent automatic insertion of the http:// scheme into the visible link text when pasting bare hostnames or URLs without an explicit scheme.
- Ensure pasted links are validated and normalized before being applied to the document selection.

Tests:
- Add a regression test verifying that pasting a bare hostname creates a link with an http:// href while keeping the visible text unchanged.


packages/editor/src/extensions/markdown/Link/paste-plugin.test.ts
```diff
@@ -41,6 +42,20 @@ describe('link paste plugin', () => {
...
+        const state = EditorState.create({
+            schema,
+            doc: startDoc,
+            selection: TextSelection.create(startDoc, startDoc.tag.a),
+            plugins,
+        });
+        const view = new EditorView(null, {state});
+        dispatchPasteEvent(view, {'text/plain': 'ya.ru'});
+        expect(view.state.doc).toMatchNode(doc(p(lnkYa('ya.ru'))));
+        same('[ya.ru](ya.ru)', view.state.doc);
```
makhnatkin: Currently, when inserting a link, it will be inserted with the protocol. After this fix, judging by the test, there won't be any protocol.

I suggest changing it to

```
same('[ya.ru](http://ya.ru)', view.state.doc);
```

packages/editor/src/extensions/markdown/Link/linkifyRawHrefPlugin.test.ts
```diff
@@ -0,0 +1,28 @@
...
+function mdWithPlugin() {
+    const md = MarkdownIt({linkify: true});
+    linkifyRawHrefPlugin(md);
+    return md;
+}
+
+describe('linkifyRawHrefPlugin', () => {
+    it('uses raw hostname in href for pasted text with trailing words', () => {
+        const md = mdWithPlugin();
+        expect(md.renderInline('ya.ru dasda')).toBe('<a href="ya.ru">ya.ru</a> dasda');
```
makhnatkin: If I understand correctly, the `linkifyRawHrefPlugin` was added to avoid adding the protocol when the user hasn't inserted it. In my opinion, the correct approach is to always use the protocol in the link itself, as before, but not display it in the text.

So, the plugin seems unnecessary

saxumcordis: What if i paste fdqn for ssh?

makhnatkin:
@sourcery-ai review

packages/editor/src/extensions/markdown/Link/paste-plugin.ts
```diff
@@ -74,17 +76,35 @@ export function linkPasteEnhance({markupParser: parser}: ExtensionDeps) {
...
     // TODO: should we process HTML here?
     const text = data.getData(DataTransferType.Text);
     const match = parser.matchLinks(text);
     if (match?.[0]) {
-        const {raw, url} = match[0];
-        if (raw === text) return url;
+        const m = match[0];
+        const {raw} = m;
+        if (raw === text) {
+            const href = parser.normalizeLink(m.schema ? text : m.url);
```
makhnatkin: When you paste `test@ya.ru`, the link is broken: the href becomes `test@ya.ru` instead of `mailto:test@ya.ru` (before it was correct). The problem is in the line `const href = parser.normalizeLink(m.schema ? text : m.url)` — for e-mail it takes the href from the plain text `test@ya.ru`, which has no `mailto:` scheme, so the protocol is lost. It is safer to always take the href from `m.url` — this field already has the correct scheme for all cases (`ya.ru` → `http://ya.ru`, `test@ya.ru` → `mailto:test@ya.ru`, `http://ya.ru` → `http://ya.ru`), so you can just write `const href = parser.normalizeLink(m.url)` without the ternary

Also add a test for pasting an e-mail plz.

makhnatkin: The ternary `label = m.schema ? text : m.raw` is dead: we are inside the guard `if (raw === text)`, and `m.raw === raw`, so both branches are equal to `text`. It looks like the label depends on the scheme, but it does not. Let's write `const label = text;` then, together with the href fix, the function reads as "href from `m.url`, visible text is exactly what was pasted".

packages/editor/src/extensions/markdown/Link/linkifyRawHrefPlugin.test.ts
makhnatkin: A bare FQDN is recognized by linkify as an `http://` link anyway this is how it works, and it was the same before the PR.

This PR only changes the visible text: the FQDN now stays in the text as is, and the scheme goes into the href. We cannot tell "this is an ssh host, not a website" from the clipboard, so the behavior is correct for the editor.

d3m1d0v (approved):

PR #1052, opened 2026-04-09
refactor(Image,ImgSize): extend Image base extension instead of replacing it

d3m1d0v:
## Summary
- Replace `builder.addNode(...)` in `ImgSizeSpecs` with `overrideNodeSpec` / `overrideMarkdownTokenParserSpec` / `overrideNodeSerializerSpec`, eliminating duplication of the base `image` node spec (same pattern applied to `YfmHeading` previously)
- Extract `imageToMarkdown()` serializer factory to `ImageSpecs/utils.ts` so `ImgSizeSpecs` can extend it via `renderExtra` callback instead of copy-pasting the serializer logic
- Update `yfm.ts` and `yfm-specs.ts` presets so `ImageSpecs` is loaded through the base preset and `ImgSize` only overrides it; fix pre-existing bug `isNumber(width) ? height : null` → `isNumber(width) ? width : null`

### DIFF
--- a/packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/const.ts
+++ b/packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/const.ts
@@ -1,11 +1,10 @@
 import {ImsizeAttr} from '@diplodoc/transform/lib/plugins/imsize/const.js';
 
+import {ImageAttr} from 'src/extensions/markdown/Image/ImageSpecs/const';
+
 export const ImgSizeAttr = {
-    Alt: ImsizeAttr.Alt,
-    Src: ImsizeAttr.Src,
-    Title: ImsizeAttr.Title,
+    ...ImageAttr,
     Width: ImsizeAttr.Width,
     Height: ImsizeAttr.Height,
-    Loading: 'loading',
     Id: 'id',
 } as const;
--- a/packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/const.ts
+++ b/packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/const.ts
@@ -1,6 +1,8 @@
 import {ImsizeAttr} from '@diplodoc/transform/lib/plugins/imsize/const.js';
 
-import {ImageAttr} from 'src/extensions/markdown/Image/ImageSpecs/const';
+import {ImageAttr} from '../../../markdown/Image/ImageSpecs/const';
+
+export {imageNodeName} from '../../../markdown/Image/ImageSpecs/const';
 
 export const ImgSizeAttr = {
     ...ImageAttr,
--- a/packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/index.ts
+++ b/packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/index.ts
@@ -3,12 +3,12 @@ import imsize from '@diplodoc/transform/lib/plugins/imsize/index.js';
 import isNumber from 'is-number';
 import type {NodeSpec} from 'prosemirror-model';
 
-import type {ExtensionAuto} from '../../../../core';
-import {imageToMarkdown} from '../../../markdown/Image/ImageSpecs';
-import {imageNodeName} from '../../../markdown/Image/const';
+import type {ExtensionAuto} from '#core';
 
-import {ImgSizeAttr} from './const';
-import {renderImgSizeExtra} from './utils';
+import {ImageSpecs} from '../../../markdown/Image/ImageSpecs';
+
+import {ImgSizeAttr, imageNodeName} from './const';
+import {imageToMarkdown, renderImgSizeExtra} from './utils';
 
 export {ImgSizeAttr};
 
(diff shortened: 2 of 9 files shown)


packages/editor/src/extensions/yfm/ImgSize/YfmImage.test.ts
```diff
@@ -12,7 +13,7 @@ const {
     markupParser: parser,
     serializer,
 } = new ExtensionsManager({
-    extensions: (builder) => builder.use(BaseSchema, {}).use(ImgSizeSpecs, {}),
+    extensions: (builder) => builder.use(BaseSchema, {}).use(ImageSpecs).use(ImgSizeSpecs, {}),
```
makhnatkin: The test had to be updated because `overrideNodeSpec` now requires the node to be registered beforehand. In practice, this is unlikely to affect real usage since most cases go through presets. However, this effectively changes the public contract and looks like a breaking change — `ImgSizeSpecs` is no longer self-contained.

Instead of adapting the test to this behavior, I would suggest a small plan to keep the extension self-contained and avoid leaking this requirement to consumers.

First, add a helper method to `ExtensionBuilder.ts`:

~~~ts
// Returns true if a node with the given name is already registered.
hasNode(name: string): boolean {
    return Boolean(this.#nodeSpecs[name] || this.#rawNodeSpecs[name]);
}
~~~

Then update `ImgSizeSpecs/index.ts` so it can safely auto-load its dependency when needed:

~~~ts
export const ImgSizeSpecs: ExtensionAuto<ImgSizeSpecsOptions> = (builder, opts) => {
    // MAJOR: remove implicit dependency on ImageSpecs
    // Auto-load ImageSpecs when used outside of a preset
    if (!builder.hasNode(imageNodeName)) {
        builder.use(ImageSpecs);
    }

    // ...overrides
};
~~~

Finally, the test can be reverted back to its original version, since the extension should handle this internally.

packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/utils.ts
```diff
@@ -0,0 +1,11 @@
+import type {Node} from '#pm/model';
+
+import type {ImgSizeAttr} from './const';
+
+export function renderImgSizeExtra(_state: unknown, node: Node): string {
```
makhnatkin: nit:
`_state: unknown` works fine here since the parameter is unused, but the expected type is already know. Why not be explicit?

packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/index.ts
```diff
@@ -31,91 +23,51 @@ export const ImgSizeSpecs: ExtensionAuto<ImgSizeSpecsOptions> = (builder, opts)
...
 
-                        return {
-                            [ImgSizeAttr.Src]: (dom as Element).getAttribute(ImgSizeAttr.Src),
-                            [ImgSizeAttr.Alt]: (dom as Element).getAttribute(ImgSizeAttr.Alt),
-                            [ImgSizeAttr.Title]: (dom as Element).getAttribute(ImgSizeAttr.Title),
-                            [ImgSizeAttr.Loading]: (dom as Element).getAttribute(
-                                ImgSizeAttr.Loading,
-                            ),
-                            [ImgSizeAttr.Height]: isNumber(height) ? height : null,
-                            [ImgSizeAttr.Width]: isNumber(width) ? height : null,
```
makhnatkin, line 64: The bug in `parseDOM.getAttrs` went unnoticed because it wasn’t actually covered by tests. It probably makes sense to add a small regression test for this case:

~~~ts
import {DOMParser} from 'prosemirror-model';

it('should correctly parse width and height from HTML img element', () => {
    const dom = document.createElement('div');
    dom.innerHTML = '<img src="img.png" width="200" height="100">';

    const parsed = DOMParser.fromSchema(schema).parse(dom);
    const imgNode = parsed.firstChild!.firstChild!;

    expect(imgNode.attrs[ImgSizeAttr.Width]).toBe('200');
    expect(imgNode.attrs[ImgSizeAttr.Height]).toBe('100');
});
~~~

packages/editor/src/extensions/yfm/ImgSize/ImgSizeSpecs/const.ts
```diff
@@ -1,11 +1,10 @@
 import {ImsizeAttr} from '@diplodoc/transform/lib/plugins/imsize/const.js';
 
+import {ImageAttr} from 'src/extensions/markdown/Image/ImageSpecs/const';
```
makhnatkin: nit:

Two files in the same folder, two different import styles:

```typescript
// ImgSizeSpecs/const.ts imports directly from sub-module
import {ImageAttr} from 'src/extensions/markdown/Image/ImageSpecs/const';

// ImgSizeSpecs/index.ts imports from index
import {imageToMarkdown} from '../../../markdown/Image/ImageSpecs';
```
`ImageAttr` is already re-exported from the index, so `const.ts` can use the same approach:

```typescript
import {ImageAttr} from 'src/extensions/markdown/Image/ImageSpecs';
```

packages/editor/src/extensions/yfm/ImgSize/YfmImage.test.ts
d3m1d0v: Good idea, I'll add new methods to builder in a separate PR

d3m1d0v: #1063

makhnatkin (approved):

PR #1054, opened 2026-04-13
test: updated circular-deps check configuration

makhnatkin:
1. Fixed the condition. `parseInt('0', 10) || 99` evaluates to 99 and threshold of 0 was never reachable. 
2. Added the flag to count only actual runtime cycles. Without `transform: true`, dpdm treats `import type` as runtime dependencies, reporting 34 phantom cycles (47 total vs 13 real).

### DIFF
--- a/packages/editor/scripts/check-circular-deps.js
+++ b/packages/editor/scripts/check-circular-deps.js
@@ -3,10 +3,14 @@ const process = require('process');
 
 const {parseDependencyTree, parseCircular, prettyCircular} = require('dpdm');
 
-const threshold = parseInt(process.argv[2], 10) || 99; // Default to 99 if no argument provided
+const threshold = process.argv[2] !== undefined ? parseInt(process.argv[2], 10) : 0;
+if (isNaN(threshold)) {
+    console.error(`Invalid threshold argument: "${process.argv[2]}"`);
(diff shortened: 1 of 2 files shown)


packages/editor/scripts/check-circular-deps.js
```diff
@@ -3,10 +3,14 @@ const process = require('process');
...
-const threshold = parseInt(process.argv[2], 10) || 99; // Default to 99 if no argument provided
+const threshold = process.argv[2] !== undefined ? parseInt(process.argv[2], 10) : 0;
+if (isNaN(threshold)) {
+    console.error(`Invalid threshold argument: "${process.argv[2]}"`);
+    process.exit(1);
+}
 
 parseDependencyTree('./src/index.ts', {
-    /* options, see https://github.com/acrazing/dpdm?tab=readme-ov-file#api-reference */
+    transform: true,
```
makhnatkin, line 13: With `transform: true`, dpdm ignores dependencies used only for types, so it shows only real cycles that affect runtime behavior

see https://github.com/acrazing/dpdm?tab=readme-ov-file#options

Suggest fixing runtime circular dependencies first, and addressing type-only circular dependencies later.

d3m1d0v (approved):

PR #1057, opened 2026-04-13
fix(markup): fix circular deps

makhnatkin:
Two cycle categories:

**Category D** — `files-upload-facet.ts` passed `FilesUploadPlugin.extension` to `Facet.define({enables})`.
CodeMirror evaluates `enables` at module load time, not call time. `FilesUploadPlugin` imported
`FileUploadHandlerFacet` — both files deadlocked at startup.
Fix: remove `enables`, register `FilesUploadPlugin.extension` explicitly in `create.ts`.

**Category A** — `commands/inline.ts` imported `DirectiveSyntaxFacet` through the `../codemirror` barrel.
Fix: import directly from `../codemirror/directive-facet`.

### DIFF
--- a/packages/editor/src/markup/codemirror/files-upload-facet.ts
+++ b/packages/editor/src/markup/codemirror/files-upload-facet.ts
@@ -2,15 +2,12 @@ import {Facet} from '@codemirror/state';
 
 import type {FileUploadHandler} from '../../utils/upload';
 
-import {FilesUploadPlugin} from './files-upload-plugin';
-
 export type {FileUploadHandler};
 
 export const FileUploadHandlerFacet = Facet.define<
     {fn: FileUploadHandler; imageWithDimensions?: boolean; enableNewImageSizeCalculation?: boolean},
     {fn: FileUploadHandler; imageWithDimensions?: boolean; enableNewImageSizeCalculation?: boolean}
 >({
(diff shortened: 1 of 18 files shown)


packages/editor/src/markup/codemirror/files-upload-facet.ts
```diff
@@ -2,15 +2,12 @@ import {Facet} from '@codemirror/state';
...
 
-import {FilesUploadPlugin} from './files-upload-plugin';
-
 export type {FileUploadHandler};
 
 export const FileUploadHandlerFacet = Facet.define<
     {fn: FileUploadHandler; imageWithDimensions?: boolean; enableNewImageSizeCalculation?: boolean},
     {fn: FileUploadHandler; imageWithDimensions?: boolean; enableNewImageSizeCalculation?: boolean}
 >({
-    enables: FilesUploadPlugin.extension,
```
makhnatkin, line 13: FYI:

Before, the plugin was not enabled automatically either. It was enabled only when `FileUploadHandlerFacet.of(...)` was added, inside `if (params.uploadHandler)`.

Now the logic is almost the same. The only difference is that the plugin is added explicitly in `create.ts`.

d3m1d0v (approved):

PR #1061, opened 2026-04-13
fix(bundle): fix circular deps

makhnatkin:
Category C — `bundle/MarkdownEditorView.tsx` declared and exported `cnEditorComponent`.
`bundle/HorizontalDrag.tsx` imported `cnEditorComponent` from `MarkdownEditorView`; but
`MarkdownEditorView` imported the `HorizontalDrag` component — closing the cycle.

Fix: extract `cnEditorComponent` to `bundle/editor-classname.ts` (which imports nothing from
within `bundle/`). Both `MarkdownEditorView.tsx` and `HorizontalDrag.tsx` import from `editor-classname.ts`.

After this PR: **0 circular dependencies**. Threshold set to 0.


packages/editor/src/bundle/MarkdownEditorView.tsx
```diff
@@ -25,16 +25,16 @@ import {MarkupEditorView} from './MarkupEditorView';
...
+import {cnEditorComponent} from './editor-classname';
 import {EditorSettings, type EditorSettingsProps, type SettingItems} from './settings';
 import {stickyCn} from './sticky';
 import type {MToolbarData, MToolbarItemData, WToolbarData, WToolbarItemData} from './toolbar/types';
 import {getToolbarsConfigs} from './toolbar/utils/toolbarsConfigs';
 import type {MarkdownEditorMode} from './types';
 
+import './MarkdownEditorView.scss';
 import '../styles/styles.scss';
-import './MarkdownEditorView.scss'; // eslint-disable-line import/order
```
d3m1d0v, line 35: It is very likely that this is an important order of importing styles and should not be changed unless necessary.

d3m1d0v (approved):

PR #1069, opened 2026-04-16
build: move gulp tasks and deps to new gulp-tasks package

d3m1d0v:
## Description

- All gulp tasks for building packages have been moved to a separate package `gulp-tasks` located in `infra/`
- `gulp-tasks` exports a `registerBuildTasks()` function and re-exports the `gulp` package
- `markdown-editor` package use `gulp-tasks` in their `gulpfile` to build themselves
- `gulp-cli` has been added to the default catalog in the pnpm workspace


infra/gulp-tasks/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "build"
+  ],
+  "exports": {
+    ".": "./index.js",
+    "./build": "./build.js"
+  },
+  "dependencies": {
+    "@gravity-ui/gulp-utils": "1.0.3",
+    "@types/gulp": "4.0.18",
+    "@types/gulp-sass": "6.0.0",
```
makhnatkin, line 17: `@types/gulp` and `@types/gulp-sass` are only needed for development and build-time typing. Should we move them to `devDependencies`?

infra/gulp-tasks/build.js
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Registers build tasks for the project using Gulp.
+ * @param {object} config Configuration object
+ * @param {string} config.version Version string to embed in the build
+ * @param {string} config.buildDir Root build output directory
+ * @param {string} config.nodeModulesDir Path to node_modules directory
+ */
+export function registerBuildTasks({version, buildDir, nodeModulesDir}) {
+    const BUILD_DIR = buildDir;
```
makhnatkin, line 27: nit: 
Assigning a parameter to a new constant looks unnecessary here.

infra/gulp-tasks/build.js
```diff
@@ -0,0 +1,110 @@
...
+    const BUILD_DIR_CJS = path.resolve(BUILD_DIR, 'cjs');
+    const BUILD_DIR_ESM = path.resolve(BUILD_DIR, 'esm');
+    const NODE_MODULES_DIR = nodeModulesDir;
```
makhnatkin, line 30: nit:
These values do not really feel like constants in the usual sense. Would it make sense to use camelCase instead?

infra/gulp-tasks/package.json
d3m1d0v: this is an infra package, it is used to build packages in `packages/`.
that's why I don't think it's necessary

d3m1d0v: `gulp-tasks` package re-exports the gulp package, so types from `@types/gulp` should be considered an important dependency

infra/gulp-tasks/build.js
d3m1d0v: yeah. I wanted to minimize changes to source code, so I used variable names that were in the original `gulpfile.js`

infra/gulp-tasks/build.js
d3m1d0v: yeah. I wanted to minimize changes to source code, so I used variable names that were in the original `gulpfile.js`

makhnatkin (approved):

PR #1074, opened 2026-04-20
feat: added page-constructor plugin

separatrixxx:
Added the `YfmPageConstructor` plugin. `@diplodoc/page-constructor-extension` is used for block rendering.

The plugin allows you to correctly display and edit PC blocks.


demo/tests/visual-tests/__snapshots__/YfmExtensions.visual.test.tsx-snapshots/Extensions-YFM-YFM-HTML-dark-chromium-linux.png
makhnatkin, line 1: @separatrixxx 

In this case, you suggest replacing the correct screenshot with an incorrect one, this is incorrect.

makhnatkin (changes requested):

PR #1132, opened 2026-05-12
feat(YfmTable): support header rows in yfm-table

d3m1d0v:
## Summary by Sourcery

Add support for configurable header rows in YFM tables, including parsing/serialization, editor controls, and visual behavior, and update related tooling and tests.

New Features:
- Allow YFM tables to declare header rows via a header-rows attribute that is parsed from markup and rendered back to markup and DOM.
- Expose a header rows option in the YfmTable extension and demos to enable header row controls in the editor UI, including toggling rows as headers from the row menu.

Bug Fixes:
- Align YfmTabs ARIA ids and aria-controls values by prefixing generated panel ids consistently for regular and radio-style tabs.

Enhancements:
- Decorate header rows in tables with data-header attributes and columnheader roles for improved visuals and accessibility, and prevent drag-and-drop operations that would break header sections.
- Adjust table row insertion, deletion, and drag-and-drop commands to keep the header-rows count in sync when header rows are added, removed, or moved.
- Refactor non-structural transaction detection into a shared utility and reuse it in the folding plugin to avoid unnecessary recomputation.
- Extend table focus, DnD styling, and cell querying logic to support both th and td elements as first-column controls.

Build:
- Bump @diplodoc/transform dependency to 4.75.1 in the workspace.

Tests:
- Add unit and visual tests covering header-rows parsing/serialization, header toggling behavior, and its interaction with row insertion, deletion, and drag-and-drop.
- Add Playwright helpers and stories to exercise header row behavior in the table DnD demo, and adjust the Playwright Docker script and package scripts for running tests with UI.


makhnatkin:
1. I suggest not using a switcher. Instead, change the icon and text: “Enable header” / “Disable header”.
2. I suggest disabling drag for enabled headers.
3. I would not add a header toggle for column selection, because users may think it affects the whole column, not only the first cell.

d3m1d0v:
> 2\. I suggest disabling drag for enabled headers.

Done

d3m1d0v:
> 3\. I would not add a header toggle for column selection, because users may think it affects the whole column, not only the first cell.

Done

packages/editor/src/extensions/yfm/YfmTable/plugins/YfmTableControls/components/FloatingMenuControl/FloatingMenuControl.tsx
```diff
@@ -46,58 +51,89 @@ export const FloatingMenuControl: React.FC<FloatingMenuControlProps> =
         onInsertAfterClick,
         onRemoveRangeClick,
         onRemoveTableClick,
+        canMakeRowHeader,
```
makhnatkin: I suggest replacing these four props:

```
canMakeRowHeader: boolean;
canUnsetRowHeader: boolean;
onMakeRowHeader: () => void;
onUnsetRowHeader: () => void;
```
With just two props: one for the state and one for the handler.

For example:
```
isRowHeader: boolean;
onRowHeaderChange: (value: boolean) => void;
```

makhnatkin (approved):

PR #1140, opened 2026-06-08
feat: add `insert()` method to editor API

ReFFaT:
I added the isertAt method, which adds to the cursor if the parameter is not passed, and to the index if it is passed. If the index is out of range, it uses the standard isert and prepend methods.

## Summary by Sourcery

Add support for inserting markup at a specific position or at the current cursor across editor implementations.

New Features:
- Expose an insertAt(markup, index?) API on the public Markdown editor and underlying editors to insert markup at the cursor or at a given absolute position.

Documentation:
- Document the new insertAt method on the public editor and content handler interfaces.

Tests:
- Add unit tests covering insertAt behavior for negative indices (prepend), in-range indices, out-of-range indices (append), and default cursor-based insertion in both Wysiwyg and markup editors.

### DIFF
--- a/packages/editor/src/common/index.ts
+++ b/packages/editor/src/common/index.ts
@@ -13,9 +13,5 @@ export interface ContentHandler {
     prepend(markup: MarkupString): void;
     append(markup: MarkupString): void;
     moveCursor(position: 'start' | 'end'): void;
-    /**
-     * Insert markup at the current cursor position.
-     * If `index` is provided, insert at that position instead.
-     */
-    insertAt(markup: MarkupString, index?: number): void;
+    insert(markup: MarkupString): void;
 }
--- a/packages/editor/src/common/index.ts
+++ b/packages/editor/src/common/index.ts
@@ -13,4 +13,9 @@ export interface ContentHandler {
     prepend(markup: MarkupString): void;
     append(markup: MarkupString): void;
     moveCursor(position: 'start' | 'end'): void;
+    /**
+     * Insert markup at the current cursor position.
+     * If `index` is provided, insert at that position instead.
+     */
+    insertAt(markup: MarkupString, index?: number): void;
 }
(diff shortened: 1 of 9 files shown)


packages/editor/src/common/index.ts
```diff
@@ -13,4 +13,9 @@ export interface ContentHandler {
     prepend(markup: MarkupString): void;
     append(markup: MarkupString): void;
     moveCursor(position: 'start' | 'end'): void;
+    /**
+     * Insert markup at the current cursor position.
+     * If `index` is provided, insert at that position instead.
+     */
+    insertAt(markup: MarkupString, index?: number): void;
```
d3m1d0v: Let's rename `insertAt() -> insert()` to be consistent with other methods.

And remove second `index` argument, as it works differently in different editor mods.

d3m1d0v (commented):
And fix issues from sourcery

packages/editor/src/core/ContentHandler.test.ts
```diff
@@ -158,4 +164,34 @@ describe('WysiwygContentHandler', () => {
         });
         expect(fn).toThrow();
     });
+
+    it('should insert inline at current cursor position', () => {
```
d3m1d0v, line 173: plz add test cases with replacement behavior

packages/editor/src/markup/editor.test.ts
```diff
@@ -96,4 +96,14 @@ describe('MarkupContentHandler', () => {
         });
         expect(fn).toThrow();
     });
+
+    it('should insert at current cursor position', () => {
```
d3m1d0v, line 100: plz add test cases with replacement behavior

packages/editor/src/core/ContentHandler.test.ts
d3m1d0v: also, add test cases with block insertions (e.g. quote)

d3m1d0v (approved):

PR #1166, opened 2026-06-30
feat: add scroll container prop

DaffPunks:
If editor placed in scrollable container, this happens:

https://github.com/user-attachments/assets/5229b7bf-6068-4600-8a4d-0cfa70178931

Because isStickyActive depends on window.addEventListener('scroll', ...). When window is not scrollable, toolbar don't add class: `g-md-editor-sticky_sticky-active`

Untitled - Frame 1 (5)

Now, with scrollContainerRef if behaves like this:
https://preview.gravity-ui.com/md-editor/1166/?path=/story/experiments-sticky-toolbar-in-scroll-container--story

---

## Summary by Sourcery

Add support for sticky toolbars that work inside a custom scroll container instead of only the browser window.

New Features:
- Introduce an optional scrollContainerRef prop across editor and toolbar views to enable sticky behavior within a specified scrollable container.
- Add a demo Storybook story showcasing a markdown editor with a sticky toolbar inside a scrollable container.

Enhancements:
- Extend the useSticky hook to listen to scroll-related events on a provided container element and compute stickiness relative to that container.
- Adjust MarkdownEditorView styling to handle layouts when a custom scroll container is used via a dedicated modifier class.


packages/editor/src/react-utils/useSticky.ts
```diff
@@ -36,9 +57,15 @@ export function useSticky<T extends HTMLElement>(elemRef: React.RefObject<T>) {
         function observe() {
             rafId = null;
             if (!elemRef.current) return;
-            const refPageOffset = elemRef.current.getBoundingClientRect().top;
-            const stickyOffset = parseInt(getComputedStyle(elemRef.current).top, 10);
-            const stickyActive = refPageOffset <= stickyOffset;
+            const stickyOffset = parseInt(getComputedStyle(elemRef.current).top);
```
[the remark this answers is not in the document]
DaffPunks: 1. In a real-world scenario (an editor with a fixed toolbar), the content above the toolbar doesn't change dynamically after mounting.
2. Refs to stable DOM nodes don't change during the component's lifetime. If the container is remounted, the entire tree is remounted, and the hook is recreated. Rebinding listeners when the ref changes is like engineering a non-existent case.

packages/editor/src/react-utils/useSticky.ts
```diff
@@ -4,25 +4,46 @@ import {useEffectOnce, useLatest} from 'react-use';
...
     useEffectOnce(() => {
         let rafId: number | null = null;
+        const scrollContainer = scrollContainerRef?.current ?? null;
+
+        const naturalTopFromContainer =
+            scrollContainer && elemRef.current
+                ? elemRef.current.getBoundingClientRect().top -
+                  scrollContainer.getBoundingClientRect().top -
+                  scrollContainer.clientTop
+                : null;
```
d3m1d0v, line 30: I think it's better to move this to the `observe()`

demo/src/stories/experiments/sticky-toolbar/StickyToolbar.stories.tsx
d3m1d0v, line 1: Please, move this story to `demo/src/stories/examples`

PR #1186, opened 2026-07-21
feat(ImgSize): add ability to override relative image URLs

separatrixxx:
Added `resolveImageSrc` hook. 

The host application may store files outside the browser-accessible filesystem (for example, VSCode webview). The relative path `./img.png` is not resolved there; the image is not displayed. The callback allows the host to convert `src` to a displayable URL without changing the Markdown source. `resolveImageSrc` only affects the displayed `src` in ``. Node attributes and serialization in markup
use the original path. `ImgSettingsButton` and the serializer read `node.attrs`, so the original path is not distorted.

## Summary by Sourcery

Introduce configurable image src resolution for the ImgSize extension to support non-browser-accessible file locations.

New Features:
- Add resolveImageSrc option to the ImgSize extension to allow hosts to transform image src values before rendering.

Enhancements:
- Propagate extension options into ImgSize node views via ImgSizeNodeViewOpts, enabling custom image src resolution without changing stored node attributes.

### DIFF
--- a/demo/src/stories/examples/img-resolve-src/ImgResolveSrc.stories.tsx
+++ b/demo/src/stories/examples/img-resolve-src/ImgResolveSrc.stories.tsx
@@ -1,0 +1,13 @@
+import type {StoryObj} from '@storybook/react';
+
+import {ImgResolveSrcDemo as component} from './ImgResolveSrc';
+
+export const Story: StoryObj<typeof component> = {
+    args: {},
+};
+Story.storyName = 'Resolve Image Src';
+
+export default {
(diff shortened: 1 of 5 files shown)


makhnatkin:
@separatrixxx Hi! We need to add a new story for the image plugin and create a WYSIWYG demo. Please also add a Diplodoc parameter so we can see the markup mode in the preview.

makhnatkin (commented):
add demo plz

separatrixxx:
@makhnatkin done

makhnatkin (approved):

PR #1188, opened 2026-07-23
fix(Table): insert markdown tables from plain text

makhnatkin:
#### Markdown table paste from plain text

- Added plain text handling for Markdown table paste before HTML handling.
- Limited table paste handling to pipe table candidates parsed as a single table.
- Fixed row loss when pasted clipboard data contains both HTML and plain text.


packages/editor/src/extensions/markdown/Table/plugins/markdownTablePastePlugin.ts
```diff
@@ -0,0 +1,55 @@
...
+                    return true;
+                },
+            },
+        },
+    });
+
+export function isPipedMarkdownTable(text: string): boolean {
+    const trimmed = text.trim();
+    return trimmed.startsWith('|') && trimmed.endsWith('|');
+}
```
d3m1d0v: This function doesn't check whether the text is a Markdown table.
It only checks that the text begins and ends with a pipe character (|).
It doesn't check anything else.

Therefore, calling it `isPipedMarkdownTable` is incorrect.

packages/editor/src/extensions/markdown/Table/plugins/markdownTablePastePlugin.ts
```diff
@@ -0,0 +1,55 @@
...
+function shouldHandlePaste(view: EditorView, clipboardData: DataTransfer): boolean {
+    return !clipboardData.types.includes(DataTransferType.Yfm) && !isInsideCode(view.state);
+}
+
+function parsePipedMarkdownTable(text: string, parser: Parser): Slice | null {
+    if (!isPipedMarkdownTable(text)) return null;
+
+    try {
+        const content = parser.parse(text).content;
+        return content.firstChild?.type.name === TableNode.Table ? new Slice(content, 0, 0) : null;
```
d3m1d0v: It's worth checking that the table is the only child in the parsed content.

d3m1d0v (approved):

d3m1d0v (approved):

PR #1208, opened 2026-08-05
fix(markup): keep Opt+Shift+A typable on macOS

stasmarkin:
## Problem

In markup mode on macOS, <kbd>Opt</kbd>+<kbd>Shift</kbd>+<kbd>A</kbd> inserts `<!--  -->` instead of typing a character. On the US layout that combination produces `Å`, so the character is simply not typable in the editor.

The binding comes from `defaultKeymap` in `@codemirror/commands`:

```js
{ key: "Alt-A", run: toggleBlockComment },
```

It is not shadowed by accident of key naming — macOS reports the composed character in `event.key`, so `w3c-keyname` deliberately falls back to the key code for Alt combinations:

```js
var brokenModifierNames = mac || chrome && +chrome[1] < 57;
var ignoreKey = brokenModifierNames && (event.ctrlKey || event.altKey || event.metaKey) || ...
var name = (!ignoreKey && event.key) || (event.shiftKey ? shift : base)[event.keyCode] || ...
```

`keyName` therefore returns `"A"`, `runHandlers` looks up `"Alt-A"` and hits the binding directly. That is by design — it is what makes Alt combinations bindable on macOS at all, and the CodeMirror maintainer has declined to make it configurable ([codemirror/dev#1621](https://github.com/codemirror/dev/issues/1621), [#1655](https://github.com/codemirror/dev/issues/1655)).

Consumers cannot work around it either: `keymaps` is spread *after* `defaultKeymap` into the same `keymap.of([...])`, and same-key bindings run in array order, so the default always wins.

## Fix

Skip that one combination on macOS.

Worth noting that upstream already applies this convention everywhere else — every other Alt+letter binding in `defaultKeymap` carries a mac override (`Alt-l` → `Ctrl-l`, `Alt-ArrowLeft` → `Ctrl-ArrowLeft`). `Alt-A` is the only one left without, which looks like an oversight rather than a decision.

Reported upstream as [codemirror/dev#1734](https://code.haverbeke.berlin/codemirror/dev/issues/1734). Once that lands, this filter turns into a no-op and should be removed.

The filter compares the key rather than the command, and resolves it the same way `buildKeymap` does (`b[platform] || b.key`), so a binding that upstream later gives a mac override is left alone.

Other platforms are untouched: `Alt+Shift+A` is not a printable character there, and the binding matches VS Code.

Nothing is lost on macOS either — `Mod-/` (`toggleComment`, also part of `defaultKeymap`) produces exactly the same `<!--  -->`, because markdown has no line comment syntax and `toggleComment` falls back to block comments.

## Test

`keymap-macos.test.ts` stubs `navigator.platform` before the module graph is loaded (both `@codemirror/view` and `w3c-keyname` read it once, at evaluation time) and dispatches the real macOS event shape (`key: 'Å'`, `keyCode: 65`, `altKey`, `shiftKey`).

Verified it reproduces the bug on `main`:

```
● markup editor keymap on macOS › should leave Opt+Shift+A to the browser

    Expected: "text"
    Received: "<!--  -->text"
```

and passes with the fix. Full unit suite (86 suites / 817 tests), `typecheck`, eslint and prettier are clean.

## Summary by Sourcery

Skip the CodeMirror Alt-A block comment keybinding on macOS in markup mode so Opt+Shift+A remains a typable character, and add coverage to ensure the macOS keymap defers this combo to the browser.

Bug Fixes:
- Prevent Opt+Shift+A on macOS from triggering block comment insertion in the markup editor so the Å character can be typed.

Tests:
- Add a macOS-specific keymap test that stubs navigator.platform and verifies Opt+Shift+A no longer triggers the Alt-A block comment binding.

### DIFF
--- a/packages/editor/src/markup/codemirror/create.ts
+++ b/packages/editor/src/markup/codemirror/create.ts
@@ -28,6 +28,7 @@ import {Action as A, formatter as f} from '../../shortcuts';
 import type {Receiver} from '../../utils';
 import {DataTransferType, shouldSkipHtmlConversion} from '../../utils/clipboard';
 import type {DirectiveSyntaxContext} from '../../utils/directive';
+import {isMac} from '../../utils/platform';
 import type {ParseInsertedUrlAsImage} from '../../utils/upload';
 import {
     insertEmptyRow,
@@ -171,7 +172,19 @@ export function createCodemirror(params: CreateCodemirrorParams) {
                 shift: insertNewlineKeepIndent,
             },
             indentWithTab,
-            ...defaultKeymap,
+            // Temporary workaround, to be dropped once CodeMirror gives Alt-A a mac override
+            // upstream (tracker: https://code.haverbeke.berlin/codemirror/dev/issues).
+            //
+            // Opt+Shift+A types a printable character on macOS ("Å" on the US layout), and key
+            // resolution there goes by key code for Alt combinations, so a binding on it shadows
+            // the character. Upstream gives every other Alt+letter binding a mac override
+            // (Alt-l -> Ctrl-l, Alt-ArrowLeft -> Ctrl-ArrowLeft); Alt-A, bound to
+            // toggleBlockComment, is the only one left without. Its command stays reachable via
+            // Mod-/, since markdown has no line comment syntax and toggleComment then falls back
+            // to block comments.
+            ...defaultKeymap.filter(
+                (binding) => !(isMac() && (binding.mac || binding.key) === 'Alt-A'),
+            ),
             ...(disabledExtensions.history ? [] : historyKeymap),
             ...keymaps,
         ]),
--- a/packages/editor/src/markup/codemirror/keymap-macos.test.ts
+++ b/packages/editor/src/markup/codemirror/keymap-macos.test.ts
@@ -1,0 +1,66 @@
+import type {EditorView} from '@codemirror/view';
+
+import type {ReactRenderStorage as ReactRenderStorageClass} from '../../extensions';
+import type {Logger2 as Logger2Class} from '../../logger';
+import type {DirectiveSyntaxContext as DirectiveSyntaxContextClass} from '../../utils/directive';
+
+import type {createCodemirror as createCodemirrorFn} from './create';
+
+// @codemirror/view and w3c-keyname both read navigator.platform once, at module evaluation time,
+// to decide how a key combination is resolved. It has to be stubbed before the module graph is
+// loaded, hence the explicit requires below instead of top-level imports.
+Object.defineProperty(window.navigator, 'platform', {value: 'MacIntel', configurable: true});
+
+/* eslint-disable @typescript-eslint/no-var-requires */
+const ReactRenderStorage: typeof ReactRenderStorageClass =
+    require('../../extensions').ReactRenderStorage;
+const Logger2: typeof Logger2Class = require('../../logger').Logger2;
+const DirectiveSyntaxContext: typeof DirectiveSyntaxContextClass =
+    require('../../utils/directive').DirectiveSyntaxContext;
+
+const createCodemirror: typeof createCodemirrorFn = require('./create').createCodemirror;
+/* eslint-enable @typescript-eslint/no-var-requires */
+
+function createView(doc: string): EditorView {
+    return createCodemirror({
+        doc,
+        placeholder: '',
+        logger: new Logger2(),
+        onCancel: () => undefined,
+        onSubmit: () => undefined,
+        onChange: () => undefined,
+        onDocChange: () => undefined,
+        onScroll: () => undefined,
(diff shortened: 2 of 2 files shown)


stasmarkin:
To be explicit about the nature of this change: **it is a temporary workaround, not where the bug belongs.**

The root cause is in `@codemirror/commands`. Its `defaultKeymap` gives every Alt+letter binding a mac override, because those combinations type printable characters on macOS:

```js
{ key: "Alt-ArrowLeft",  mac: "Ctrl-ArrowLeft",  run: cursorSyntaxLeft }
{ key: "Alt-ArrowRight", mac: "Ctrl-ArrowRight", run: cursorSyntaxRight }
{ key: "Alt-l",          mac: "Ctrl-l",          run: selectLine }         // Alt-l types "¬"
{ key: "Alt-A",                                  run: toggleBlockComment }  // Alt-A types "Å"
```

`Alt-A` is the only one left without one (checked against `@codemirror/commands` 6.10.4), which looks like an oversight — the same rationale the CodeMirror maintainer used when declining to make macOS Alt bindings configurable in [codemirror/dev#1621](https://github.com/codemirror/dev/issues/1621) and [#1655](https://github.com/codemirror/dev/issues/1655) applies to this binding too.

Once upstream adds `mac:` to that binding, this filter becomes a no-op and should be removed. The code comment says so and points at the tracker.

One practical note: the CodeMirror issue tracker has moved off GitHub — `codemirror/dev` is archived and now lives at https://code.haverbeke.berlin/codemirror/dev/issues, where registration is currently disabled. So the upstream report is not filed yet, and there is no ETA on a fix. That is the reason for landing the workaround rather than waiting.

stasmarkin:
@sourcery-ai thanks — I've considered both, and I'd rather keep the current shape. Reasoning below, happy to be argued out of it.

**1. Hoisting the filtered `defaultKeymap` into a module-level constant**

That would move the `isMac()` call from editor construction to module evaluation, and `isMac()` is not safe to freeze:

```ts
export const isMac = (): boolean =>
    typeof navigator === 'undefined' ? false : /Mac/.test(navigator.platform);
```

If the module is first evaluated where `navigator` is undefined — SSR, or any Node-side import of the package — the constant would be computed as "not macOS" and stay that way for the process. Evaluating it per `createCodemirror` call keeps the check where it is always answered on the client.

On the cost side: `defaultKeymap` is 57 bindings, so this is one 57-element `.filter` per editor instance. The same array literal it lives in already resolves 18 shortcuts through `f.toCM()` on every construction, and construction itself builds a full `EditorState` and DOM view. The filter isn't measurable next to that, and editors aren't created in a loop.

**2. Restoring `navigator.platform` in `afterAll`**

Jest gives every test file its own `testEnvironment` instance, so the stub cannot leak into other files — and this file's stub has to stay in effect for its entire lifetime by design.

The reason the stub sits above the `require` calls rather than in a `beforeAll` is that both `@codemirror/view` and `w3c-keyname` read `navigator.platform` exactly once, at module evaluation, and derive module-level constants (`browser.mac`, `brokenModifierNames`) from it. Once those modules are loaded, restoring the property changes nothing about key resolution. An `afterAll` teardown would therefore suggest an isolation boundary that doesn't exist — a future test added to this file would still run against a macOS-resolved keymap whether or not the property was restored.

If more platforms need covering later, the correct move is a separate test file per platform, not teardown in this one.

stasmarkin:
I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en

d3m1d0v (approved):

d3m1d0v:
> One practical note: the CodeMirror issue tracker has moved off GitHub — `codemirror/dev` is archived and now lives at https://code.haverbeke.berlin/codemirror/dev/issues, where registration is currently disabled. So the upstream report is not filed yet, and there is no ETA on a fix. That is the reason for landing the workaround rather than waiting.

You can use your GitHub account to log into his repository to create issue: https://code.haverbeke.berlin/user/login

stasmarkin:
Upstream report filed: [codemirror/dev#1734 — *defaultKeymap binds Alt-A on macOS, where it shadows a printable character*](https://code.haverbeke.berlin/codemirror/dev/issues/1734).

For anyone following the thread: the CodeMirror tracker has moved off GitHub, but signing in there with GitHub OAuth works, so the report is on the real tracker rather than the forum.

The issue leads with the precedent I missed when opening this PR — codemirror/dev#474 and #486 are the same bug for `Alt-l` → `selectLine`, which blocked `ł` on the Polish layout and `¬` on UK English. That one was patched the same day it was reported and shipped in @codemirror/commands 0.18.2 as the `mac: "Ctrl-l"` override that is still there today. `Alt-A` looks like the case that was missed in that pass, which makes an upstream fix fairly likely.

Nothing changes for this PR in the meantime — if the override lands upstream, the filter here becomes a no-op and can be dropped.

PR #1238, opened 2026-09-16
feat(core): add builder methods to register node and mark views

d3m1d0v:
## Summary by Sourcery

Enable explicit registration and construction of node and mark views through the extension builder.

New Features:
- Add builder methods for registering node and mark view factories independently of entity definitions.
- Build editor node and mark views with finalized extension dependencies while preserving registered views over plugin-provided views.

Enhancements:
- Preserve view registrations across schema, parser, serializer, and repeated build operations while preventing duplicate or conflicting registrations.

Tests:
- Add coverage for view registration validation, immutability, dependency construction, factory timing, and editor view precedence.


packages/editor/src/core/ExtensionBuilder.ts
makhnatkin, line 226: nit: right now the order of calls matters for views — `addNodeView('x', v).addNodeSpec('x', s)` fails, while the other way around works.

I think we can improve this. It seems to be needed only because `addNode` at `:226` does `views[name] = base.view` even when `base.view` is not set, overwriting a view added earlier. With `views[name] = base.view ?? views[name]` (and a throw if both are set), both orders should work.

d3m1d0v: > nit: right now the order of calls matters for views — addNodeView('x', v).addNodeSpec('x', s) fails, while the other way around works.

That’s exactly how I intended it

d3m1d0v: The order requirement is intentional: a custom view extends an already registered node or mark. I’d prefer to keep spec registration first and fail early when the entity is missing. With this contract, `addNode` cannot overwrite a previously registered view.

makhnatkin (approved):

PR #1245, opened 2026-09-16
refactor(YfmHtmlBlock): use new builder methods to register nodes

d3m1d0v:
## Summary by Sourcery

Migrate YFM HTML block registration to the newer builder methods while preserving its node view integration.

Enhancements:
- Refactor the YFM HTML block extension to register node specs, Markdown parsing and serialization independently through the builder APIs.
- Register the HTML block node view explicitly after loading its specs and mark the legacy nodeView option as deprecated.


packages/editor/src/extensions/additional/YfmHtmlBlock/YfmHtmlBlockSpecs/index.tsx
```diff
@@ -29,56 +30,56 @@ const YfmHtmlBlockSpecsExtension: ExtensionAuto<YfmHtmlBlockSpecsOptions> = (
...
-            },
-            toMd: (state, node) => {
-                // The parser includes the line break before closing ::: in srcdoc.
-                // Drop only that structural newline; any remaining trailing newline is content.
-                const srcdoc = String(
-                    node.attrs[YfmHtmlBlockConsts.NodeAttrs.srcdoc] || '',
-                ).replace(/\n$/, '');
+            toDOM: (node) => ['iframe', node.attrs],
+        }))
+        .addMarkdownTokenParserSpec('yfm_html_block', () => ({
```
makhnatkin: `YfmHtmlBlockConsts.NodeName` is `'yfm_html_block'` and it is used on the next line and in `addNodeSpec` above. Using a raw string only here is inconsistent, and a typo would not be caught by types - it would just break parsing at runtime.

makhnatkin, line 45: nit: `YfmHtmlBlockConsts.NodeName` is `'yfm_html_block'` and is already used on the next line - could we pass it here too? A typo in a raw string fails neither the types nor the build, it just silently stops the parsing.

makhnatkin (approved):

packages/editor/src/extensions/additional/YfmHtmlBlock/YfmHtmlBlockSpecs/index.tsx
d3m1d0v: Token names and node names can differ and should not be defined by a shared variable.

PR #1246, opened 2026-09-16
refactor(Mermaid): use new builder methods to register nodes

d3m1d0v:
## Summary by Sourcery

Refactor Mermaid node registration to use the builder’s dedicated specification and node-view APIs.

Enhancements:
- Refactor Mermaid extension registration to use separate builder methods for node specs, Markdown token parsing, serialization, and node views.

Documentation:
- Mark the Mermaid nodeView option as deprecated in favor of registering the view through the builder.


packages/editor/src/extensions/additional/Mermaid/MermaidSpecs/index.tsx
```diff
@@ -7,51 +7,51 @@ import {MermaidConsts, defaultMermaidEntityId, mermaidNodeName} from './const';
...
-                dnd: {props: {offset: [8, 1]}},
+            parseDOM: [],
+            toDOM(node) {
+                return ['div', node.attrs];
             },
-            toMd: (state, node) => {
-                const content: string = node.attrs[MermaidConsts.NodeAttrs.content] || '';
+            dnd: {props: {offset: [8, 1]}},
+        }))
+        .addMarkdownTokenParserSpec('mermaid', () => ({
```
makhnatkin, line 33: nit:

`mermaidNodeName` is imported in this file and is used on the next line and in `addNodeSpec` above. You can use it here too instead of the `'mermaid'` string - a typo in the string is not checked by types and would silently break parsing.

makhnatkin (approved):

packages/editor/src/extensions/additional/Mermaid/MermaidSpecs/index.tsx
d3m1d0v: Token names and node names can differ and should not be defined by a shared variable.

PR #1248, opened 2026-09-16
fix(core): keep parser aliases in their entity pipeline

d3m1d0v:
Fixed a builder bug where extra parser tokens for nodes also created marks, and vice versa. This prevents schema errors when multiple tokens use the same node or mark. Added regression tests without changing the public API.

## Summary by Sourcery

Keep parser aliases in their owning entity pipeline and prevent cross-registration between nodes and marks.

Bug Fixes:
- Keep parser aliases confined to their node or mark pipeline, preventing aliases from being incorrectly registered as the opposite entity type and avoiding schema errors.

Documentation:
- Clarify that Docker and Podman requirements apply specifically to visual Playwright tests.

Tests:
- Add regression coverage for ignored parser aliases and multiple node or mark aliases, including end-to-end parsing and serialization checks.


packages/editor/src/core/ExtensionBuilder.ts
```diff
@@ -111,7 +112,9 @@ function resolveParserPipeline(
 
             if (primaryParserToken[entityName]) {
                 // Extra parser-only token targeting an existing entity
-                extraTokenNames.push(entry.tokenName);
+                if (entityNames.has(entityName)) {
```
makhnatkin: This stops the alias from leaking into the *other* pipeline, but it still becomes a schema entry in its own one. `ExtensionsManager.processNode` passes every `nodes()` entry to `SchemaSpecRegistry.addNode`, and a parser-only entry carries `spec: {}` — so after this fix `schema.marks.fence` is gone (the new test asserts exactly that), while `schema.nodes.fence` is still there: an empty leaf node named after a markdown token. Mark aliases behave the same way on the mark side.

#1240 takes a different route: it drops `extraTokenNames` and both `buildParserOnly*Entry` helpers, and such tokens go to the parser registry through a separate `parserOnlyTokens()`. They never reach the schema, and the node/mark ambiguity this guard protects against stops existing.

Both are fine fixes, but they conflict in `ExtensionBuilder.ts` (checked with `git merge-tree`, not by eye), so whichever lands second has to be rewritten. Could we agree on the order? If this one is needed sooner, it can go first and I'll rebase #1240 on top of it — otherwise this guard is probably not needed at all.

makhnatkin (approved):

PR #1252, opened 2026-09-17
fix: preserve sticky toolbar styling at fractional offsets

makhnatkin:
#### Sticky toolbar precision

- Fixed missing toolbar backgrounds and click layers caused by fractional offsets and subpixel rounding.
- Added regression coverage for scrolling, resized offsets, and toolbar interactions in both editor modes.

##### What was broken

The sticky top is `calc(var(--g-md-toolbar-sticky-offset, 0px) + 8px)`. If the host sets the offset to `0.5px`, the computed top is `8.5px`. The element really sticks at `8.5`, but `parseInt` read the offset as `8`, so `8.5 <= 8` was false and the active class was never set. That class holds both the `z-index` and the `::before` background, so the toolbar had no background and clicks went to the content under it.

##### Why `+ 1`

The browser lays out in 1/64 of a CSS pixel. At a zoom other than 100% the rect top is `8.015625` instead of `8`, so the check flips on and off at the edge. One pixel of slack stops that. The toolbar turns on one pixel earlier, which is not visible. This is a heuristic, not an exact fix: an exact one would compare rounded 1/64 units, or use an `IntersectionObserver` sentinel.


makhnatkin:
Superseded by #1250: same fix (parseFloat + 1px tolerance), with a named constant and the tests from here ported in a trimmed form.

